PR #61, opened 2021-12-09
Edit and add signing limits

richardiphillips:


src/Service.Host/client/src/actions/employeesActions.js
```diff
@@ -0,0 +1,15 @@
+// import FetchApiActions from './FetchApiActions';
```
lewisrenfrew, line 1: commented out thing

richardiphillips: ah fixed it already but stashed it. will sort next time.

lewisrenfrew (approved):

PR #63, opened 2021-12-10
Part supplier ut 3

lewisrenfrew:
- mainly  pr'ing to see if any of my client weirdness happens on sys


richardiphillips (approved):

richardiphillips:
i've added a new abstract method to the facade service in main so you might get a bit of merging pain.

For your facade services = probably easiest to throw away the changes in main completely and then just let resharper re-add the missing method.

PR #64, opened 2021-12-10
Orders by supplier (v0) 

iaincrawf:
Just need to work out why domain test isn't getting any rows, will step through tomorrow 

Doesn't include most of the report options that the existing Oracle one has, will add them in next pr


richardiphillips:
I haven't looked at the code yet but I'm guessing from what you're saying about format that you're probably returning the resultsModel without converting to the ReportReturnResource? 

If so you just need to apply an extension that's in the shared library.

src/Persistence.LinnApps/Repositories/PurchaseOrderRepository.cs
```diff
@@ -0,0 +1,40 @@
...
+        public PurchaseOrderRepository(ServiceDbContext serviceDbContext)
+            : base(serviceDbContext.PurchaseOrders)
+        {
+            this.serviceDbContext = serviceDbContext;
+        }
+
+        public override IQueryable<PurchaseOrder> FilterBy(Expression<Func<PurchaseOrder, bool>> expression)
+        {
+            return this.serviceDbContext.PurchaseOrders.Where(expression).Include(o => o.Details)
+                .ThenInclude(d => d.PurchaseDelivery).AsNoTracking().AsQueryable();
```
lewisrenfrew, line 25: do we need AsQueryable() here? (Not saying we don't just curious)

src/Service.Host/client/src/components/reports/OrdersBySupplierOptions.js
```diff
@@ -0,0 +1,120 @@
...
+                        Run Report
+                    </Button>
+                </Grid>
+            </Grid>
+        </Page>
+    );
+}
+
+OrdersBySupplierReportOptions.propTypes = {
+    history: PropTypes.shape({ push: PropTypes.func }).isRequired
```
lewisrenfrew, line 117: history isn't passed in as a prop so I imagine this PropTypes is doing nada

src/Service.Host/client/src/index.js
```diff
@@ -22,7 +24,9 @@ const render = Component => {
             <ThemeProvider theme={linnTheme}>
                 <SnackbarProvider dense maxSnack={5}>
                     <AppContainer>
-                        <Component store={store} />
+                        <LocalizationProvider dateAdapter={AdapterDateMoment} locale="de-DE">
```
lewisrenfrew, line 27: local de-DE? what does that do?

tests/Integration/Integration.Tests/PurchaseOrdersReportModuleTests/WhenGettingOrdersBySupplier.cs
```diff
@@ -0,0 +1,98 @@
...
+                                   };
+
+            reportReturnResource.ReportResults.Add(reportResult);
+
+            this.FacadeService
+                .GetOrdersBySupplierReport(Arg.Any<OrdersBySupplierSearchResource>(), Arg.Any<IEnumerable<string>>())
+                .Returns(new SuccessResult<ReportReturnResource>(reportReturnResource));
+            var resource = new OrdersBySupplierSearchResource { From = "2/11/21", To = "2/12/21" };
+
+            var builder = new UriBuilder("http://localhost:51699/purchasing/reports/orders-by-supplier/report");
```
lewisrenfrew, line 55: I think I managed to get query params in here in a (slightly) neater way if you look at some of my SupplierModuleTests

src/Persistence.LinnApps/Repositories/PurchaseOrderRepository.cs
iaincrawf: Looks like we don't when I take it out, will test! Do you happen to see what I've done wrong re integration test getting no data? It feels handler-y but not sure

src/Service.Host/client/src/index.js
iaincrawf: Ahh I meant to post about this, the date was in American format in datepickers and I couldn't get it to accept en-GB sort of thing, german was at least the right way round just with .s ie 16.12.21

iaincrawf: I'll post in it dev actually, definitely only a temporary thing

src/Service.Host/client/src/components/reports/OrdersBySupplierOptions.js
iaincrawf: Yeah will we ever have props now? Do we ditch propTypes completely?

richardiphillips: We'll still have to set PropTypes for all Objects we new up such a subcomponents etc. We just don't have as many now because we don't always have the "child" component to the container each time

lewisrenfrew (approved):

PR #67, opened 2021-12-13
Fix launch settings

lewisrenfrew:
missed a } before


richardiphillips (approved):
what happened to these. I thought they were ok before?

lewisrenfrew:
> what happened to these. I thought they were ok before?

I accidentally changed them in a previous PR. Although I definitely didn't manually change them so I think my Visual Studio did something to them

PR #70, opened 2021-12-16
Logging

richardiphillips:
As usual classes ending with 2 are destined for common. 

This was the least invasive and most understandable solution I came up with. 

It leaves a bit for the user to do i.e. make the repository call but apart from that it leaves it very flexible in terms of what you can log. Can just be a note field, a single operation or whatever. 

I did come up with some much more complicated versions that hid more of the work but was difficult to not put restrictions on the generic types or add complication that is just unnecessary overhead when you're not logging.


src/Facade/Services/FacadeResourceService2.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        protected abstract T CreateFromResource(TResource resource);
+
+        protected abstract void UpdateFromResource(T entity, TUpdateResource updateResource);
+
+        protected abstract Expression<Func<T, bool>> SearchExpression(string searchTerm);
+
+        protected abstract void DeleteOrObsoleteResource(T entity);
+
+        protected abstract void SaveToLogTable(string actionType, int userNumber, T entity, TResource resource, TUpdateResource updateResource);
```
iaincrawf, line 151: I'm not sure if you guys have spoken about this before or if it wouldn't work, but depending how often logging happens would we ever make this (and possibly the delete too) virtual methods instead of abstract, so we don't have to override them with not implemented if we weren't using them in that facade service?

iaincrawf (approved):
Cool, do we have a lot of log tables? I've not come across many so far but I like the sound of them especially for important stuff

src/Facade/Services/FacadeResourceService2.cs
richardiphillips: that's not a bad idea. They're abstract to advertise that they're there really but perhaps the lesser used ones could be less in your face

lewisrenfrew (approved):

src/Facade/Services/FacadeResourceService2.cs
lewisrenfrew: I prefer abstract fwiw. virtual would mean you'd have to provide a default implementation in the generic service which would probably lead you back towards the complications. Of course you could just throw a NotImplementedException in the default implemetation, but I think thats unclear as virtual implies 'I have a default implementation you can use'

richardiphillips: yeah. Also a good point. It's more dangerous having 'hidden' code that actually just throws an exception

iaincrawf: Yeah I did think it would be less obvious they were there, pretty much extreme laziness on my part thinking about writing a few lines less per facade 😄 Happy to leave as they are or up to you if you want to try one and see how we like it, thought it'd be worth asking and seeing others' opinions on it. I do get Lewis' point that it implies there's a default implementation, in practice I'd hope it's not too different from having our own not implemented(s) though - hopefully we'd just get an error and remember we have to implement bits

PR #72, opened 2021-12-17
add delete to grid rows

richardiphillips:
implement delete in client and add Lewis's edit allowed icon


iaincrawf (approved):
Noice. Ross doing react testing has reminded me that we should probably start adding some front end tests in purchasing too, I haven't done any yet but will add basic ones to the report(s) at some point

richardiphillips:
> Noice. Ross doing react testing has reminded me that we should probably start adding some front end tests in purchasing too, I haven't done any yet but will add basic ones to the report(s) at some point

yep. me too

PR #74, opened 2021-12-22
Orders by supplier v1

iaincrawf:
all of the filtering options

Still comparing results from each of the options to the old report to make sure they're right but getting late so thought I'd put it up for pr (won't merge until I've checked it's all right)


src/Domain.LinnApps/PurchaseOrderDetails.cs
```diff
@@ -12,10 +14,10 @@ public class PurchaseOrderDetail
 
         public string PartNumber { get; set; }
 
-        public PurchaseOrder PurchaseOrder { get; set; }
-
         public PurchaseOrderDelivery PurchaseDelivery { get; set; }
```
richardiphillips, line 17: there can be multiple deliveries per order line. I don't know if this matters for the report but would be wrong if this becomes the actual model. (if it is then it should be in the purchase orders folder)

src/Domain.LinnApps/Reports/IPurchaseOrdersReportService.cs
```diff
@@ -8,6 +8,6 @@ public interface IPurchaseOrdersReportService
     {
         ResultsModel GetOrdersByPartReport(DateTime from, DateTime to, string partNumber);
 
-        ResultsModel GetOrdersBySupplierReport(DateTime from, DateTime to, int supplierId);
+        ResultsModel GetOrdersBySupplierReport(DateTime from, DateTime to, int supplierId, bool includeReturns, bool outstandingOnly, bool includeCancelled, string includeCredits, string stockControlled);
```
richardiphillips, line 11: line goes off the end of my screen

tests/Unit/Domain.LinnApps.Tests/OrdersBySupplierReportServiceTests/WhenGettingReport.cs
```diff
@@ -84,10 +83,17 @@ public void SetUp()
             this.SupplierRepository.FindById(Arg.Any<int>())
                 .Returns(new Supplier { Name = "We sell stuff", SupplierId = this.supplierId });
 
+            //this.PurchaseOrdersPack.OrderIsCompleteSql(Arg.Any<int>(), Arg.Any<int>()).Returns(true);
```
richardiphillips, line 86: comment

tests/Unit/Domain.LinnApps.Tests/OrdersBySupplierReportServiceTests/WhenGettingReport.cs
```diff
@@ -20,7 +20,6 @@ public class WhenGettingReport : ContextBase
         private readonly int supplierId = 77282;
```
richardiphillips, line 20: typo in namespace

tests/Unit/Domain.LinnApps.Tests/OrdersBySupplierReportServiceTests/WhenGettingReport.cs
```diff
@@ -101,7 +107,8 @@ public void ShouldCallRepos()
         [Test]
         public void ShouldReturnData()
         {
-            this.results.ReportTitle.DisplayValue.Should().Be($"Purchase Orders By Supplier - {this.supplierId}: We sell stuff");
+            this.results.ReportTitle.DisplayValue.Should()
+                .Be($"Purchase Orders By Supplier - {this.supplierId}: We sell stuff");
             this.results.Rows.Count().Should().Be(1);
             var row = this.results.Rows.First();
             row.RowId.Should().Be($"{this.orderNumber}/1");
```
richardiphillips, line 114: the results model has got a few public methods like getgridvalue and getgridtextvalue which are useful for testing that reports are generated correctly

richardiphillips (approved):

src/Domain.LinnApps/PurchaseOrderDetails.cs
iaincrawf: Ah yeah, good point. I think I'll need to rewrite a few lines slightly and work out if it's a delivery per row of report or if a row will use sum of delivery qtys etc. Think possibly a row per delivery because it takes dates from the delivery. I'll note this + to use the nice testing methods into an issue for myself when I'm back

PR #75, opened 2021-12-23
Part supplier create working 

lewisrenfrew:
- Anything with a 2 in the type name will be moved back to common before merge


src/Persistence.LinnApps/Repositories/PartSupplierRepository.cs
```diff
@@ -45,7 +45,8 @@ public IQueryable<PartSupplier> FindAll()
 
         public void Add(PartSupplier entity)
         {
-            throw new NotImplementedException();
+            this.serviceDbContext.Entry(entity.Part).State = EntityState.Unchanged;
```
lewisrenfrew, line 48: db tries to add a new part Entity instead of tracking the existing without this line. Will try and find a better fix before I merge

richardiphillips: Feels like there must be a way to include part number in the key without implicating the whole part itself but if not adding a bridgeId wouldn't be the worst thing in the world. 

Also i've made the Add method overridable now so you can just inherit entityframeworkrepository rather then irepository

src/Facade/ResultHandler2.cs
```diff
@@ -0,0 +1,42 @@
...
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    using Linn.Common.Facade;
+    using Linn.Common.Facade.Carter;
+
+    using Microsoft.AspNetCore.Http;
+
+    public abstract class ResultHandler2<T> : IHandler
```
richardiphillips, line 12: what's changed in the incubated classes?

richardiphillips (approved):

src/Facade/ResultHandler2.cs
richardiphillips: ah just saw the common pr. That's cool

src/Persistence.LinnApps/Repositories/PartSupplierRepository.cs
lewisrenfrew: I'll still need all my .Include()'s in this repositorys FindBy() etc methods. Is there some way to get them using the entityframeworkrepository too?

richardiphillips: yes you can just override the methods that you need to

lewisrenfrew: of course! thanks. Also changing the PartRepository call to a FindById() stopped EF core trying to insert a new Part... which makes sense since how else would it know the part in question was actually a part that already existed without knowing the Id 🙃

PR #78, opened 2022-01-14
example of how to get stuff from redux store into components for testing

lewisrenfrew:
- you can just mock out the redux state that useSelector will act on as shown


iaincrawf (approved):
Cool

richardiphillips (approved):
if it works, this is huge!

RossStewart:
yeah, definitely something for show & tell!

PR #91, opened 2022-01-25
Orders by supplier export + tests

iaincrawf:
+ support multiple purchase order deliveries on a purchase order (detail)


src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
```diff
@@ -52,5 +67,32 @@ private async Task GetOrdersBySupplierReport(HttpRequest req, HttpResponse res)
...
+
+            this.csvStreamWriter.WriteModel(results);
+
+            var contentDisposition = new ContentDisposition
+                                         {
+                                             FileName =
+                                                 $"ordersBySupplier{resource.From.Substring(0, 10)}_To_{resource.To.Substring(0, 10)}.csv"
+                                         };
+
+            this.stream.Position = 0;
```
iaincrawf, line 94: I think this could also be replaced with something like 
` res.Headers.Add("Content-Range", "bytes */*");`
Or whatever represents all bytes from what you were saying @lewisrenfrew, I don't have a preference

src/Service.Host/client/src/components/__tests__/reports/ordersBySupplierOptions.test.js
```diff
@@ -0,0 +1,52 @@
...
+        expect(screen.getByLabelText('Stock Controlled')).toHaveDisplayValue('All');
+
+        expect(screen.getByLabelText('Include Credits')).toBeInTheDocument();
+        expect(screen.getByLabelText('Include Credits')).toHaveDisplayValue('No');
+
+        expect(screen.getByLabelText('Include Cancelled')).toBeInTheDocument();
+        expect(screen.getByLabelText('Include Cancelled')).toHaveDisplayValue('No');
+
+        expect(screen.getByText('Run Report')).toBeInTheDocument();
+    });
```
lewisrenfrew, line 51: could try a test to make sure the right thing happens when you click to run report

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
```diff
@@ -11,23 +14,35 @@
...
+    using Microsoft.Extensions.FileProviders;
     using Microsoft.Extensions.Primitives;
 
     public class PurchaseOrdersReportModule : CarterModule
     {
         private readonly IPurchaseOrderReportFacadeService purchaseOrderReportFacadeService;
 
+        private readonly CsvStreamWriter csvStreamWriter;
+
+        private MemoryStream stream;
```
lewisrenfrew, line 36: I would declare a

lewisrenfrew: stream in a using block

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
```diff
@@ -11,23 +14,35 @@
...
+
         public PurchaseOrdersReportModule(IPurchaseOrderReportFacadeService purchaseOrderReportFacadeService)
         {
             this.purchaseOrderReportFacadeService = purchaseOrderReportFacadeService;
             this.Get("/purchasing/reports/orders-by-supplier", this.GetApp);
             this.Get("/purchasing/reports/orders-by-supplier/report", this.GetOrdersBySupplierReport);
+            this.Get("/purchasing/reports/orders-by-supplier/export", this.GetOrdersBySupplierExport);
+            this.stream = new MemoryStream();
+            this.csvStreamWriter = new CsvStreamWriter(this.stream);
```
lewisrenfrew, line 45: Could this not come in via dependency injection?

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
lewisrenfrew: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
```diff
@@ -52,5 +67,32 @@ private async Task GetOrdersBySupplierReport(HttpRequest req, HttpResponse res)
...
+
+            resource.Returns = req.Query.As<string>("Returns");
+            resource.Outstanding = req.Query.As<string>("Outstanding");
+            resource.Cancelled = req.Query.As<string>("Cancelled");
+            resource.Credits = req.Query.As<string>("Credits");
+            resource.StockControlled = req.Query.As<string>("StockControlled");
+
+            var results = this.purchaseOrderReportFacadeService.GetOrdersBySupplierExport(resource, req.HttpContext.GetPrivileges());
+
+            this.csvStreamWriter.WriteModel(results);
```
lewisrenfrew, line 86: It does feel a bit wrong to be doing the IEnumerable<IEnumerable<string>> -> csv stream conversion in the Carter Module but I can't think of a perfect solution off of the top of my head.

Maybe we could have a further Response extension method like FromStreamToCsv in common that makes use of the existing FromStream() one and does the csv convert too

lewisrenfrew (approved):

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
lewisrenfrew: I mean the StreamWriter, not the Stream. I'd declare the stream at the point of use inside a using block as per above comment.

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
iaincrawf: something like this? I'm not sure I can use dependency injection for the csvStreamWriter because I need to give it the stream when I new it up

`  using (MemoryStream stream = new MemoryStream())
            {
                CsvStreamWriter csvStreamWriter = new CsvStreamWriter(stream);
                csvStreamWriter.WriteModel(results);
                var contentDisposition = new ContentDisposition
                                             {
                                                 FileName =
                                                     $"ordersBySupplier{resource.From.Substring(0, 10)}_To_{resource.To.Substring(0, 10)}.csv"
                                             };

                this.stream.Position = 0;
                await res.FromStream(this.stream, "text/csv", contentDisposition);
            }`

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
iaincrawf: I don't mind it too much since it's just calling common, I guess we could rewrite stuff to do it in the facade, and then I guess we'd return a stream from the facade? Not sure that's nice either

src/Service.Host/client/src/components/__tests__/reports/ordersBySupplierOptions.test.js
iaincrawf: Haha this was my original plan. I couldn't get the button by role so ended up just leaving it, thought was maybe overkill for a setup report test. I'll have another quick look and see, maybe I'll be able to fire a click event at what it's got by text actually

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
richardiphillips: i'd +1 for the facade or domain service definitely rather than the module.

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
lewisrenfrew: That'll work. I think as of C# 8 you can omit the curly brackets and just delcare it like a normal var, e.g.

using var stream = new MemoryStream(); 

Werdly my visual studio complains when I do that, but its definitely legal syntax and it compiles

lewisrenfrew: except this.stream.Position = 0 will need to just be stream.Position... now obviously

src/Service/Modules/Reports/PurchaseOrdersReportModule.cs
iaincrawf: Instead of returning a stream from facade, I could pass it through from the module so I can still use its contents, then within the facade (or domain even) can new up the csvStreamWriter and call WriteModel. Does that sound better to you guys?

lewisrenfrew: I guess we could just have a method in the FacadeService that returns a stream as opposed to a list of lists of strings

iaincrawf: Tried returning the stream, but because we're doing using within the facade now it fails. I guess I'll pass it in and return void or a success result
image

iaincrawf: Will fix after lunch 😫 
image

iaincrawf: still getting that ^, tempted to revert and try work out a facade solution that works in my next pr

PR #92, opened 2022-01-26
Preferred supplier ut tests + loads of missing form logic

lewisrenfrew:


richardiphillips (approved):
skimmed it a bit but looks good as far as i can see

PR #93, opened 2022-01-27
Ords by part v0

iaincrawf:
- Waiting to hear back from Stewart on whether they need "to part" function (haven't done it in this, might need to have a think if they do want it)
- Haven't done any js tests yet but will this morning
- Part search isn't picking up everything that it should, not sure if it's my proxy setting's wrong or if I've done something else. Will investigate


src/Service.Host/client/src/components/Root.js
```diff
@@ -71,6 +73,16 @@ const Root = ({ store }) => (
...
                                     component={OrdersBySupplier}
                                 />
+                                <Route
+                                    exact
+                                    path="/purchasing/reports/orders-by-part"
+                                    component={OrdersByPartOptions}
+                                />
+                                <Route
+                                    exact
+                                    path="/purchasing/reports/orders-by-part/:id"
```
lewisrenfrew, line 83: How come you need the id in the path here? Is it a part number string? If so it could have forward slashes in it and cause some awkwardness

iaincrawf: You know I wondered that as I typed/copied it but wasn't awake enough to fully think it through, it should be /report I'm sure. Will change!

tests/Unit/Facade.Tests/PurchaseOrdersReportServiceTests/ContextBase.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Linn.Purchasing.Facade.Tests.ResourceBuilderTests.OrdersBySupplierReportServiceTests
+﻿namespace Linn.Purchasing.Facade.Tests.ResourceBuilderTests.PurchaseOrdersReportServiceTests
```
lewisrenfrew, line 1: Is this a ResourceBuilder test? Looks like the Sut in this context is the PurchaseOrderReportFacadeService

lewisrenfrew (approved):

tests/Unit/Facade.Tests/PurchaseOrdersReportServiceTests/ContextBase.cs
lewisrenfrew: (just the namespace that looks wrong I mean, not the actual test)

iaincrawf: was just facade service tests yeah, have amended the namespace! Good spot. I guess these will also indirectly test the resource builder but just took the style of tests from stores

lewisrenfrew: They do in this case, but they probably shouldn't since these are unit tests. I'd advocate for seperate unit tests for the builders, and then mocking them out in here

PR #97, opened 2022-01-28
More part suppliers ut

lewisrenfrew:


RossStewart (approved):
Looks good. Really like the testing layout.

PR #104, opened 2022-02-11
Supplier ut 4

lewisrenfrew:
- Where' Tab (4/7)
- bonus address utility. I've made this a standalone at /purchasing/addresses but you can also render it in a modal as a pop up from another component
- no front end tests yet but I'm gonna wait until the state of the ui is fixed before I do any tests to avoid rewrites


src/IoC/ServiceExtensions.cs
```diff
@@ -72,8 +74,13 @@ public static IServiceCollection AddFacades(this IServiceCollection services)
                 .AddTransient<IFacadeResourceService<PriceChangeReason, string, PriceChangeReasonResource, PriceChangeReasonResource>, PriceChangeReasonService>()
                 .AddTransient<IFacadeResourceService<PartCategory, string, PartCategoryResource, PartCategoryResource>, PartCategoriesService>()
                 .AddTransient<IPurchaseOrderReportFacadeService, PurchaseOrderReportFacadeService>()
+<<<<<<< HEAD
```
richardiphillips, line 77: mergey business

richardiphillips (approved):

PR #105, opened 2022-02-11
Report tweaks

iaincrawf:
tweaks to sort feedback below:
•	When you go to find the supplier would be good to see the supplier codes beside the supplier name.
•	If no advised date, a date is defaulting to Jan 0001 – should be blank
•	No total value showing at bottom of page which is handy to have.
•	Don’t think the Non Stock Controlled filter is working properly.  Shows stock items.
•	The currency, if not GBP, is showing the currency value (which is actually the same on the current report).  Do we want to leave like this or should it be in GBP as all other values are in GBP in Oracle?  Not a real issue either way as long as it is clear the value is the native currency.

I still haven't sorted the dates being American format thing but it seems like they haven't noticed it yet


src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -73,14 +73,20 @@ public ResultsModel GetOrdersByPartReport(DateTime from, DateTime to, string par
...
                 }
             }
 
             reportLayout.SetGridData(values);
             var model = reportLayout.GetResultsModel();
 
+            foreach ((var row, int i) in model.Rows.Select((value, i) => (value, i)))
+            {
+                model.SetGridValue(i, 5, model.GetGridValue(i, 5), decimalPlaces: 2);
+                model.SetGridValue(i, 7, model.GetGridValue(i, 7), decimalPlaces: 2);
```
iaincrawf, line 87: setting decimal places bit I mentioned in standup

src/Domain.LinnApps/PurchaseOrders/PurchaseOrder.cs
```diff
@@ -24,5 +24,7 @@ public class PurchaseOrder
         public string Overbook { get; set; }
 
         public decimal? OverbookQty { get; set; }
+
+        public string CurrencyCode { get; set; }
```
lewisrenfrew, line 28: You could make this a Currency as opposed to just a string since it'll probably need to be that eventually anyway

src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -183,15 +193,16 @@ private void AddSupplierReportColumns(SimpleGridLayout reportLayout)
...
+                        new AxisDetailsModel("NetTotalCurrency", "Net Total (Currency)", GridDisplayType.Value),
                         new AxisDetailsModel("Delivery", "Delivery", GridDisplayType.TextValue),
                         new AxisDetailsModel("Qty", "Qty", GridDisplayType.TextValue),
                         new AxisDetailsModel("ReqDate", "Req Date", GridDisplayType.TextValue),
                         new AxisDetailsModel("AdvisedDate", "Advised Date", GridDisplayType.TextValue)
                     });
         }
 
-
         private void AddPartReportColumns(SimpleGridLayout reportLayout)
```
lewisrenfrew, line 206: could probably be static

src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -201,28 +212,28 @@ private void AddPartReportColumns(SimpleGridLayout reportLayout)
...
                         new AxisDetailsModel("Currency", "Currency", GridDisplayType.TextValue),
-                        new AxisDetailsModel("NetTotal", "Net Total", GridDisplayType.TextValue),
+                        new AxisDetailsModel("NetTotalCurrency", "Net Total (Currency)", GridDisplayType.Value),
                         new AxisDetailsModel("Delivery", "Delivery", GridDisplayType.TextValue),
-                        new AxisDetailsModel("Qty", "Qty", GridDisplayType.TextValue)
+                        new AxisDetailsModel("Qty", "Qty", GridDisplayType.Value)
                     });
         }
 
         private void ExtractSupplierReportDetails(
```
lewisrenfrew, line 231: static?

src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -298,16 +327,17 @@ private void AddPartReportColumns(SimpleGridLayout reportLayout)
                     {
                         RowId = currentRowId,
                         ColumnId = "AdvisedDate",
-                        TextDisplay = delivery.DateAdvised.ToString("dd-MMM-yyyy")
+                        TextDisplay = delivery.DateAdvised?.ToString("dd-MMM-yyyy")
                     });
         }
 
         private void ExtractDetailsForPartReport(
```
lewisrenfrew, line 334: static?... you get the picture

lewisrenfrew (approved):

src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
richardiphillips: yeah doesn't feel right

src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -183,15 +193,16 @@ private void AddSupplierReportColumns(SimpleGridLayout reportLayout)
                         new AxisDetailsModel("QtyOrd", "Qty Ordered", GridDisplayType.TextValue),
                         new AxisDetailsModel("QtyRec", "Qty Rec", GridDisplayType.TextValue),
                         new AxisDetailsModel("QtyInv", "Qty Inv", GridDisplayType.TextValue),
-                        new AxisDetailsModel("NetTotal", "Net Total", GridDisplayType.TextValue),
+                        new AxisDetailsModel("BaseNetTotal", "Net Total (GBP)", GridDisplayType.Value),
+                        new AxisDetailsModel("Currency", "Currency", GridDisplayType.TextValue),
+                        new AxisDetailsModel("NetTotalCurrency", "Net Total (Currency)", GridDisplayType.Value),
```
richardiphillips, line 198: I haven't used this for a while but pretty sure you can specify the decimal places in here?

iaincrawf: I must not have actually checked the AxisDetailsModel class yesterday - you can't via the constructor, but you're right it's on the class. Have just defined the decimal columns at the top and set the decimalPlaces before adding them to the columns list

PR #106, opened 2022-02-11
Unack supp

richardiphillips:


src/Service.Host/client/src/components/Root.js
```diff
@@ -105,8 +105,8 @@ const Root = ({ store }) => (
                                 />
                                 <Route
                                     exact
-                                    path="/purchasing/reports/spend-by-supplier/report"
-                                    component={SpendBySupplier}
```
lewisrenfrew, line 109: How come this is getting overwritten?

richardiphillips: how did I manage that? Thanks. Good spot

lewisrenfrew (approved):

src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -163,18 +170,97 @@ public ResultsModel GetOrdersByPartReport(DateTime from, DateTime to, string par
...
+                }
+            }
+        }
+
+        public ResultsModel GetUnacknowledgedOrders(DateTime startDate, DateTime endDate, int supplierId, int? organisationId)
+        {
+            throw new NotImplementedException();
+        }
+
         private void AddSupplierReportColumns(SimpleGridLayout reportLayout)
```
lewisrenfrew, line 254: I said it on Iain's PR so I'll say it here for equality: these could be static. I don't know if we care really but apparently its more performant when methods that can be static are static

iaincrawf: I think this is one of the methods I've just made static so will just need merge conflicts resolved. Might need to check the files you've moved too, I think I made a change or two to PurchaseOrder.cs, PurchaseOrderDelivery and PurchaseOrderDetails and it doesn't look like it'll catch it as a merge conflict because I've made the changes to them in their old location

src/Service/Service.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Linn.Common.Configuration" Version="2.0.0" />
     <PackageReference Include="linn.common.facade.carter" Version="5.2.0" />
     <PackageReference Include="Linn.Common.Pdf" Version="1.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.0" />
```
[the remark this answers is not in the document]
richardiphillips: dunno. i didn't consciously add this so i'll try taking it out.

PR #110, opened 2022-02-16
Part supplier and suppliers bits

lewisrenfrew:
- create supplier endpoint etc
 - remove unused stuff from Part Suppliers


lewisrenfrew:
I just realised the react tests are failing since I removed some fields...

PR #112, opened 2022-02-18
set body margin to 0 to avoid screen flicker/horizontal scroll bar

iaincrawf:
margin is set by something to 8px causing horizontal scrollbar to appear, in stores is overridden but not sure what by. We could figure it out if anyone has any ideas or we can just do this and set it to 0 manually.
Stores from dev tools:
image
and purchasing with this change in dev tools:
image


richardiphillips (approved):
nice

PR #116, opened 2022-02-23
add get by supplier groups

richardiphillips:


iaincrawf (approved):
Looks good, how will purchasing actually decide which group a supplier is in, out of interest? Is it things like by a certain part type or locations or something?

lewisrenfrew (approved):

richardiphillips:
> Looks good, how will purchasing actually decide which group a supplier is in, out of interest? Is it things like by a certain part type or locations or something?

it's a field on the suppliers table. At the moment there's no front end way of setting it as it will very rarely change but might add it to Lewis's new form at some point in the future.

PR #119, opened 2022-02-25
Supplier contacts

lewisrenfrew:
- Doesn't support creating a new contact yet, only linking a supplier to an existing contact 
- I'm not sure I'm going down the right path with this but PR'ing for now


lewisrenfrew:
rewrite incoming

PR #120, opened 2022-02-25
Supplier contacts

lewisrenfrew:


src/Service.Host/client/src/itemTypes.js
```diff
@@ -120,3 +120,5 @@ export const vendorManagers = new ItemType(
 );
 
 export const planners = new ItemType('planners', 'PLANNERS', '/purchasing/suppliers/planners');
+
+export const contacts = new ItemType('contacts', 'CONTACTS', '/purchasing/suppliers/contacts');
```
richardiphillips, line 124: i would rename this supplierContacts whereever you can because contacts are still used in a lot of places and this could potentially become confusing

src/Domain.LinnApps/Suppliers/Supplier.cs
```diff
@@ -76,5 +77,7 @@ public class Supplier
         public string Notes { get; set; }
 
         public int OrganisationId { get; set; }
+
+        public IEnumerable<SupplierContact> Contacts { get; set; }
```
richardiphillips, line 81: as mentioned on the React side - i'd call these supplierContacts whereever possible

src/Facade/Services/SupplierFacadeService.cs
```diff
@@ -126,8 +127,29 @@ private static Supplier BuildEntityFromResourceHelper(SupplierResource resource)
                                         ? DateTime.Parse(resource.DateClosed) : null,
                            ReasonClosed = resource.ReasonClosed,
                            Notes = resource.Notes,
-                           OrganisationId = resource.OrganisationId
-            };
+                           OrganisationId = resource.OrganisationId,
+                           Contacts = resource.Contacts?.Select(c => new SupplierContact
+                                                                          {
+                                                                              SupplierId = resource.Id,
+                                                                              ContactId = c.Id > 0 ? c.Id : this.databaseService.GetIdSequence("CONT_SEQ"),
```
richardiphillips, line 134: i think the id should be set in the persistence layer but I can see that it in theses instances it would require the creation of whole new repository classes just for that single reason so this does seem a reasonable balance

richardiphillips (approved):

src/Service.Host/client/src/itemTypes.js
lewisrenfrew: I think this can go actually. It was a hangover from when I was allowing a lookup on existing contacts

Issue #22, opened 2021-11-19
Purchase Order Screen


richardiphillips:
PL_MINI_ORDER_ORA

Screen for raising, amending and viewing purchase orders 
Possibly change to allow multi line orders

PR #125, opened 2022-03-03
Pl credit debit notes 2

lewisrenfrew:
- I spoke to Andy and all they really need here is a simple ut to search and view/print/cancel Debit/Credit Notes, so thats all this PR does.
- He said he wants the notes to be automatically created (and the emails sent) at the point of a returns order creation, So I added a note to the Purchase Orders Screen project task #22  for whoever ends up doing that


iaincrawf (approved):
cool lgtm

PR #128, opened 2022-03-08
Ticket 15406

lewisrenfrew:


RossStewart (approved):

iaincrawf (approved):
So was close being called when should be calling cancel? The test reads to me that we're checking it closes correctly instead of now cancels correctly (but I'm also not awake yet)

lewisrenfrew:
> So was close being called when should be calling cancel? The test reads to me that we're checking it closes correctly instead of now cancels correctly (but I'm also not awake yet)

yeah, it should only call Close() in the case of Closing, but I forgot to return after calling close so it was calling Close(), Cancel() and Update() when it should have only been calling Close()

PR #131, opened 2022-03-10
Supplier ut  tests

lewisrenfrew:
- more tests. not quite tested absolutely everything but getting there

### DIFF
--- a/src/Service.Host/client/src/components/__tests__/supplierUtility.js/Supplier.test.js
+++ b/src/Service.Host/client/src/components/__tests__/supplierUtility.js/Supplier.test.js
@@ -1,0 +1,125 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from 'react';
+import { useSelector } from 'react-redux';
+import '@testing-library/jest-dom/extend-expect';
+import { screen, fireEvent } from '@testing-library/react';
+import routeData from 'react-router';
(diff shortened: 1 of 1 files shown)


richardiphillips (approved):

src/Service.Host/client/src/components/__tests__/supplierUtility.js/Supplier.test.js
```diff
@@ -0,0 +1,506 @@
...
+        useSelector.mockImplementation(callback => callback(state));
+        render(<Supplier creating />);
+
+        const nameInput = screen.getByLabelText('Phone Number');
+        fireEvent.change(nameInput, { target: { value: 'NEW SUPPLIER NAME' } });
+        const saveButton = screen.getByRole('button', { name: 'Save' });
+        fireEvent.click(saveButton);
+    });
+
+    test('shoud clear errors dispatch add action...', () => {
```
iaincrawf, line 491: doesn't matter but missing an 'L' in should here and the one above lol, also maybe an 'and', but really commenting because for most they're "should do x" but there's a few above that don't follow that like 'Page dispatch put on hold action'. Obviously doesn't stop ya putting it through but could make them all 'should blah' for the next one

iaincrawf (approved):
Noiiiice, pretty detailed

PR #133, opened 2022-03-14
Bulk leadtimes upload

lewisrenfrew:
- back end is still a little rough around the edges, but pr'ing the general client csv file upload idea.. could show and tell tomorrow if everyone is free
- supports drag n drop, and clicking to open a file browser as shown
- uses react-dropzone 
- could easily add this as a shared component if we like
Screenshot 2022-03-14 151057


src/Domain.LinnApps/PartSuppliers/PartSupplierService.cs
```diff
@@ -319,6 +319,45 @@ public PreferredSupplierChange CreatePreferredSupplierChange(PreferredSupplierCh
...
+        {
+            var successCount = 0;
+            var errors = new List<string>();
+            var leadTimeUpdateModels = changes.ToList();
+
+            foreach (var change in leadTimeUpdateModels)
+            {
+                var record = this.partSupplierRepository.FindBy(
+                    x => x.PartNumber == change.PartNumber.ToUpper().Trim()
+                         && x.SupplierRanking == 1);
```
lewisrenfrew, line 332: not sure this is right, need to check with Andy & co

richardiphillips (approved):
awesome

PR #136, opened 2022-03-16
Overbook front initial

RossStewart:
First commit, just checking more tests in and then I'll add reviewers


RossStewart:
Plan to put at least another test in and there's a few visual spacing things on the form I'd like to change but put up for review to make sure the direction is ok.

src/Service.Host/client/src/components/Overbook.js
```diff
@@ -0,0 +1,132 @@
...
+                        clearErrors();
+                        updatePurchaseOrder();
+                    }}
+                    cancelClick={() => {
+                        reduxDispatch({ type: 'initialise', payload: purchaseOrder });
+                    }}
+                    backClick={() => history.push('/purchasing/purchase-orders/allow-over-book/')}
+                />
+            </Grid>
+            <Grid item xs={12} />
```
iaincrawf, line 127: can probably delete these empty grids

src/Service.Host/client/src/components/Overbook.js
```diff
@@ -0,0 +1,132 @@
...
+import { useParams } from 'react-router-dom';
+import { makeStyles } from '@mui/styles';
+import Grid from '@mui/material/Grid';
+import Typography from '@mui/material/Typography';
+import purchaseOrderActions from '../actions/purchaseOrderActions';
+import history from '../history';
+import config from '../config';
+import purchaseOrder from '../reducers/purchaseOrder';
+
+function ViewPurchaseOrderForOverbook() {
```
iaincrawf, line 20: Since this component updates as well as view maybe could rename to AllowPurchaseOrderOverbook or something along those lines

src/Service.Host/client/src/components/Overbook.js
```diff
@@ -0,0 +1,132 @@
...
+    SaveBackCancelButtons
+} from '@linn-it/linn-form-components-library';
+import { useParams } from 'react-router-dom';
+import { makeStyles } from '@mui/styles';
+import Grid from '@mui/material/Grid';
+import Typography from '@mui/material/Typography';
+import purchaseOrderActions from '../actions/purchaseOrderActions';
+import history from '../history';
+import config from '../config';
+import purchaseOrder from '../reducers/purchaseOrder';
```
iaincrawf, line 18: I don't think the reducer will be used in the component, can check if it's used and delete if not

iaincrawf (approved):
Might want to get someone else to also review in case I've gone blind to anything but looks good to me, just a few tiny suggestions in the component

iaincrawf:
> Plan to put at least another test in and there's a few visual spacing things on the form I'd like to change but put up for review to make sure the direction is ok.

Yeah a 'get purchase order' integration test and probably good to go. Did you do facade tests last time? It looks like you've substituted the facade services in the integration tests so good to also test the facade services with their own tests if not, I think probably did them last time though

src/Service.Host/client/src/components/Overbook.js
```diff
@@ -0,0 +1,132 @@
...
+                    }}
+                    backClick={() => history.push('/purchasing/purchase-orders/allow-over-book/')}
+                />
+            </Grid>
+            <Grid item xs={12} />
+            <Grid item xs={12} />
+        </Page>
+    );
+}
+export default ViewPurchaseOrderForOverbook;
```
lewisrenfrew, line 132: Should probably add a 'Save Succesful' Message or display some kind of error message if not a success result

tests/Integration/Integration.Tests/PurchaseOrderModuletests/WhenSearchingPurchaseOrders.cs
```diff
@@ -41,13 +41,13 @@ public void SetUp()
                                           }
                                   };
 
-            this.PurchaseOrderFacadeService.FilterBy(
-                    Arg.Any<PurchaseOrderSearchResource>(),
+            this.PurchaseOrderFacadeService.Search(
+                    Arg.Any<string>(),
```
lewisrenfrew, line 45: I would put the actual string in here as opposed to an Arg.Any<string>

src/Service.Host/client/src/components/Root.js
```diff
@@ -161,6 +163,16 @@ const Root = ({ store }) => (
                                     path="/purchasing/open-debit-notes"
                                     component={OpenDebitNotes}
                                 />
+                                <Route
```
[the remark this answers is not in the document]
iaincrawf: Ah yeah I forgot about that, Ross will just have to add the routes to the purchase order module and getApp Index sort of business, right? I think they'll need to be what they are to differentiate from the future purchase order ut on the front end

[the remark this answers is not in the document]
lewisrenfrew: yep, just something to serve up the app when you hit these urls

PR #138, opened 2022-03-17
Bluerequt v0

iaincrawf:
Sorry this is a fairly large pr, got away from me. Will be small and often from now on

- basically backend + a view page for front end, lots more to be added but thought it was at the point where I really should pr
- had quite the merge with lots of conflicts in suppliers/supplier contacts/addresses but all seems to work ok, possibly worth a play on sys too though


src/Facade/Services/PurchaseOrderReqFacadeService.cs
```diff
@@ -0,0 +1,125 @@
...
+            PurchaseOrderReq entity,
+            PurchaseOrderReqResource resource,
+            PurchaseOrderReqResource updateResource)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override Expression<Func<PurchaseOrderReq, bool>> SearchExpression(string searchTerm)
+        {
+            return x => x.OrderNumber.ToString().Contains(searchTerm);
```
iaincrawf, line 69: expecting to change this search to a search on multiple things (ie filter not search) but need to check with users which fields first and haven't implemented the search page on front end yet, is a FacadeFilterResourceService in anticipation of that basically

src/Persistence.LinnApps/Repositories/PurchaseOrderReqRepository.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public PurchaseOrderReqRepository(ServiceDbContext serviceDbContext)
+            : base(serviceDbContext.PurchaseOrderReqs)
+        {
+            this.serviceDbContext = serviceDbContext;
+        }
+
+        public override PurchaseOrderReq FindById(int key)
+        {
+            var purchaseOrderReq = this.serviceDbContext.PurchaseOrderReqs
```
lewisrenfrew, line 22: could just return this straight away

src/Service/Modules/PurchaseOrderModule.cs
```diff
@@ -62,18 +66,39 @@ private readonly
...
+        private async Task GetApp(HttpRequest req, HttpResponse res)
         {
-            var result = this.currencyService.GetAll();
+            await res.Negotiate(new ViewResponse { ViewName = "Index.html" });
+        }
+
+        private async Task AllowOverbook(HttpRequest req, HttpResponse res)
+        {
+            var resource = await req.Bind<PurchaseOrderResource>();
+            resource.Privileges = req.HttpContext.GetPrivileges();
```
lewisrenfrew, line 84: no need to attach the privileges to the resource here. Just pass em straight to the facadeservice

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderReqService.cs
```diff
@@ -0,0 +1,72 @@
...
+            entity.AuthorisedBy = updatedEntity.AuthorisedBy;
+            entity.SecondAuthBy = updatedEntity.SecondAuthBy;
+            entity.FinanceCheckBy = updatedEntity.FinanceCheckBy;
+            entity.TurnedIntoOrderBy = updatedEntity.TurnedIntoOrderBy;
+            entity.NominalCode = updatedEntity.NominalCode;
+            entity.RemarksForOrder = updatedEntity.RemarksForOrder;
+            entity.InternalNotes = updatedEntity.InternalNotes;
+            entity.DepartmentCode = updatedEntity.DepartmentCode;
+
+            return true;
```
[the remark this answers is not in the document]
iaincrawf: From memory I think I thought I needed it to return something in order to say the domainservice.update returns true in my integration tests without actually calling a real method, but thinking about it void and not specifying anything should work since I've substituted it

src/Persistence.LinnApps/Repositories/PurchaseOrderReqRepository.cs
iaincrawf: Well spotted, left over from debugging so that I could see what the object it was getting from the db!

src/Service/Modules/PurchaseOrderModule.cs
iaincrawf: This is part of Ross's overbook work, I think I'll let him merge his pr into main first and then pull main into this branch again - I'm pretty sure it'll be changed on his branch

PR #140, opened 2022-03-22
Whats due in report

lewisrenfrew:


src/Domain.LinnApps/Keys/PurchaseOrderDeliveryKey.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Linn.Purchasing.Domain.LinnApps.Keys
+{
+    public class PurchaseOrderDeliveryKey
```
richardiphillips, line 3: line key? Delivery key would include delivery_seq (pl_deliveries)

lewisrenfrew: ah yeah, good catch

PR #142, opened 2022-03-23
Blue req ut v1

iaincrawf:
search page added, create properly implemented, required fields checked before save & blue_req_states pulled in now

Haven't yet emphasised some fields that are required like country typeahead etc - required doesn't seem to add the nice wee star to them at mo. If anyone's got an existing solution to that pls let me know

### DIFF
--- a/src/Facade/ResourceBuilders/PurchaseOrderReqStateResourceBuilder.cs
+++ b/src/Facade/ResourceBuilders/PurchaseOrderReqStateResourceBuilder.cs
@@ -1,0 +1,38 @@
+﻿namespace Linn.Purchasing.Facade.ResourceBuilders
+{
+    using System;
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Purchasing.Domain.LinnApps.PurchaseOrders;
+    using Linn.Purchasing.Resources;
+
+    public class PurchaseOrderReqStateResourceBuilder : IBuilder<PurchaseOrderReqState>
+    {
+        public PurchaseOrderReqStateResource Build(PurchaseOrderReqState entity, IEnumerable<string> claims)
+        {
+            if (entity == null)
+            {
+                return new PurchaseOrderReqStateResource();
+            }
+
+            return new PurchaseOrderReqStateResource
+                       {
+                           State = entity.State,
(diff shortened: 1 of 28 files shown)


src/Service.Host/client/src/components/POReqSearch.js
```diff
@@ -0,0 +1,155 @@
...
+    const handleOptionsChange = (propertyName, newValue) =>
+        setOptions({ ...options, [propertyName]: newValue });
+
+    const createUrl = utilities.getHref(applicationState, 'create');
+
+    useEffect(() => {
+        if (
+            options.reqNumber.length > 2 ||
+            options.part.length > 2 ||
+            options.supplier.length > 2
```
iaincrawf, line 64: I wanted to use useSearch for this, but couldn't get it working. This works but if folk do want to search by more than one field it's gonna fire off a request for every character they type in the second field. Once I've spoken to folk the search fields might get changed anyway though so I can come back to it

src/Facade/ResourceBuilders/PurchaseOrderReqStateResourceBuilder.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    using Linn.Common.Facade;
+    using Linn.Purchasing.Domain.LinnApps.PurchaseOrders;
+    using Linn.Purchasing.Resources;
+
+    public class PurchaseOrderReqStateResourceBuilder : IBuilder<PurchaseOrderReqState>
+    {
+        public PurchaseOrderReqStateResource Build(PurchaseOrderReqState entity, IEnumerable<string> claims)
+        {
+            if (entity == null)
```
lewisrenfrew, line 14: possibly don't need this check?

lewisrenfrew (approved):
looks good

src/Facade/ResourceBuilders/PurchaseOrderReqStateResourceBuilder.cs
iaincrawf: Yeah think you're right, I do need this sort of check in the actual Req resource builder for getting the application state I'm sure, but I don't need it here for ReqStates

PR #145, opened 2022-03-25
Overbook Log Update

RossStewart:
Using Richard's SaveToLog contribution to Common that SignalLimitLog is using in my facade service. Just double checking but it doesn't seem that the PL_OVERBOOK_ALLOWED_BY table - any idea where I'm going wrong here?

### DIFF
--- a/src/Facade/Services/PurchaseOrderFacadeService.cs
+++ b/src/Facade/Services/PurchaseOrderFacadeService.cs
@@ -13,15 +13,18 @@
         : FacadeResourceService<PurchaseOrder, int, PurchaseOrderResource, PurchaseOrderResource>
     {
         private readonly IPurchaseOrderService domainService;
+        private readonly IRepository<OverbookAllowedByLog, int> overbookAllowedByLogRepository;
 
         public PurchaseOrderFacadeService(
             IRepository<PurchaseOrder, int> repository,
             ITransactionManager transactionManager,
             IBuilder<PurchaseOrder> resourceBuilder,
-            IPurchaseOrderService domainService)
+            IPurchaseOrderService domainService,
+            IRepository<OverbookAllowedByLog, int> overbookAllowedByLogRepository)
             : base(repository, transactionManager, resourceBuilder)
         {
             this.domainService = domainService;
+            this.overbookAllowedByLogRepository = overbookAllowedByLogRepository;
         }
 
         protected override PurchaseOrder CreateFromResource(
@@ -33,12 +36,20 @@
 
         protected override void SaveToLogTable(
             string actionType, 
-            int userNumber, 
+            int userNumber,
             PurchaseOrder entity, 
             PurchaseOrderResource resource, 
             PurchaseOrderResource updateResource)
         {
-            throw new NotImplementedException();
+            var log = new OverbookAllowedByLog
+            {
+                OrderNumber = entity.OrderNumber,
+                OrderLine = 1,
+                OverbookQty = entity.OverbookQty,
+                OverbookDate = DateTime.Now,
+                OverbookGrantedBy = userNumber
+            };
+            this.overbookAllowedByLogRepository.Add(log);
         }
 
         protected override void DeleteOrObsoleteResource(
--- a/src/Facade/Services/PurchaseOrderFacadeService.cs
+++ b/src/Facade/Services/PurchaseOrderFacadeService.cs
@@ -43,6 +43,9 @@
         {
             var log = new OverbookAllowedByLog
             {
+                LogAction = actionType,
+                LogTime = DateTime.UtcNow,
+                LogUserNumber = userNumber,
                 OrderNumber = entity.OrderNumber,
                 OrderLine = 1,
                 OverbookQty = entity.OverbookQty,
--- a/src/Persistence.LinnApps/Repositories/OverbookAllowedByLogRespository.cs
+++ b/src/Persistence.LinnApps/Repositories/OverbookAllowedByLogRespository.cs
@@ -1,16 +1,8 @@
-﻿using Linn.Common.Proxy.LinnApps;
-using Linn.Purchasing.Domain.LinnApps.PurchaseOrders;
-
-namespace Linn.Purchasing.Persistence.LinnApps.Repositories
+﻿namespace Linn.Purchasing.Persistence.LinnApps.Repositories
 {
-    using System;
-    using System.Linq;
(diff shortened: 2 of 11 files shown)


src/Persistence.LinnApps/Repositories/OverbookAllowedByLogRespository.cs
```diff
@@ -0,0 +1,26 @@
...
+            : base(serviceDbContext.AllowOverbookLogs)
+        {
+            this.serviceDbContext = serviceDbContext;
+            this.databaseService = databaseService;
+        }
+
+        public override void Add(OverbookAllowedByLog entity)
+        {
+            entity.Id = this.databaseService.GetIdSequence("AN_ARBITRARY_SEQUENCE");
+            entity.LogId = entity.Id;
```
lewisrenfrew, line 22: why are there two ids?

tests/Integration/Integration.Tests/PurchaseOrderModuletests/WhenGettingPurchaseOrder.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Linn.Common.Facade;
```
lewisrenfrew, line 1: using should be inside the namespace. Also the casing on the namespace is correct but the folder has a lowercase t in tests

lewisrenfrew:
> Using Richard's SaveToLog contribution to Common that SignalLimitLog is using in my facade service. Just double checking but it doesn't seem that the PL_OVERBOOK_ALLOWED_BY table - any idea where I'm going wrong here?

do you need to do a TransactionManager commit perhaps?

src/Facade/Services/PurchaseOrderFacadeService.cs
```diff
@@ -33,12 +36,23 @@ public class PurchaseOrderFacadeService
...
             PurchaseOrderResource updateResource)
         {
-            throw new NotImplementedException();
+            var log = new OverbookAllowedByLog
+            {
+                LogAction = actionType,
+                LogTime = DateTime.UtcNow,
+                LogUserNumber = userNumber,
+                OrderNumber = entity.OrderNumber,
+                OrderLine = 1,
```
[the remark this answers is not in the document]
iaincrawf: This was part me, meant to say or ask for opinions - the allow overbook thing is for the full purchase order entity, not any specific po detail/line, so the line number in this log doesn't really mean anything now

tests/Integration/Integration.Tests/PurchaseOrderModuletests/WhenGettingPurchaseOrder.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    public class WhenGettingPurchaseOrder : ContextBase
+    {
+        private PurchaseOrder req;
+        private int orderNumberSearch;
+
+        [SetUp]
+        public void SetUp()
+        {
+            this.orderNumberSearch = 600179;
```
lewisrenfrew, line 27: would rename this var orderNumber since its not a search

src/Persistence.LinnApps/Repositories/OverbookAllowedByLogRespository.cs
RossStewart: > > Using Richard's SaveToLog contribution to Common that SignalLimitLog is using in my facade service. Just double checking but it doesn't seem that the PL_OVERBOOK_ALLOWED_BY table - any idea where I'm going wrong here?
> 
> do you need to do a TransactionManager commit perhaps?

Aye, it does remind me of that to be honest. I'll go through the SignalLog code to double check this.

iaincrawf:
> > Using Richard's SaveToLog contribution to Common that SignalLimitLog is using in my facade service. Just double checking but it doesn't seem that the PL_OVERBOOK_ALLOWED_BY table - any idea where I'm going wrong here?
> 
> do you need to do a TransactionManager commit perhaps?

I think it should do it within the update method in common, after it calls maybeSaveLog & then SaveToLogTable within that https://github.com/linn/common/blob/master/src/Linn.Common.Facade/FacadeResourceService.cs#L130

lewisrenfrew (approved):

src/Persistence.LinnApps/Repositories/OverbookAllowedByLogRespository.cs
richardiphillips: no you shouldn't have to commit. it's called before the update or add commits anyway

RossStewart: > why are there two ids?

This was something I wasn't too sure on (left it out initial commits). The entity.Id is a field within the table whilst the LogId is in the Linn.Common.Domain.LogTable abstract class, wasn't too sure if it needed to be set in the same way that the LogAction is set.

src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -509,6 +512,17 @@ private void BuildPurchaseOrderDeliveries(ModelBuilder builder)
             entity.Property(o => o.NetTotal).HasColumnName("NET_TOTAL").HasMaxLength(18);
         }
 
+        private void BuildOverbookAllowedBy(ModelBuilder builder)
+        {
+            var entity = builder.Entity<OverbookAllowedByLog>().ToTable("PL_OVERBOOK_ALLOWED_BY");
+            entity.HasKey(a => a.Id);
```
lewisrenfrew, line 518: you haven't actually told the mapper what the Id fields column name is, and you have two keys  on the OverbookAllowedByLog, Id and LogId.  I would guess that's probably why the insert isn't working

src/Facade/Services/PurchaseOrderFacadeService.cs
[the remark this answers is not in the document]
lewisrenfrew: hmm, seems like it shouldn't be on the table at all then. Could just not set the line at all here, and then drop the column from the table  once the users stop using the old allow overbooking utility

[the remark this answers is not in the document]
iaincrawf: yeah, the table has a default value of 1 which will work for now, so @RossStewart could not set it here & actually remove OrderLine from the log class & table mapping in this pr. Be good to make an issue in the purchasing project to drop the orderLine column from this log table once the old form is out of the menu, so we remember to do it

PR #149, opened 2022-03-29
fix case

lewisrenfrew:
> is it lower case for you on main? Mine has it capitalised, can't remember if I already changed it image

its lowercase. Stupid Windows doesn't really care about case but github does


iaincrawf:
is it lower case for you on main? Mine has it capitalised, can't remember if I already changed it 
image

lewisrenfrew:
> is it lower case for you on main? Mine has it capitalised, can't remember if I already changed it image

There's currently two folders on main - one with the correct case and one without - causing all kinds of confusion

iaincrawf (approved):
your changes look like it's replaced the lower case one now so hopefully that'll be it sorted

PR #150, opened 2022-03-29
Outstanding po reqs report

lewisrenfrew:


src/Service.Host/client/src/actions/index.js
```diff
@@ -162,3 +162,8 @@ export const whatsDueInReportActionTypes = makeActionTypes(
 export const purchaseOrderReqStatesActionTypes = makeActionTypes(
     itemTypes.purchaseOrderReqStates.actionType
 );
+
+export const outstandingPoReqsReportActionTypes = makeActionTypes(
```
richardiphillips, line 166: doesn't really make a difference but should be makeReportActionTypes

richardiphillips (approved):

src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -840,6 +840,8 @@ private void BuildPurchaseOrderReqs(ModelBuilder builder)
             entity.HasOne(e => e.SecondAuthBy).WithMany().HasForeignKey(x => x.SecondAuthById);
             entity.Property(e => e.Email).HasColumnName("EMAIL_ADDRESS").HasMaxLength(50);
             entity.Property(e => e.InternalNotes).HasColumnName("INTERNAL_ONLY_ORDER_NOTES").HasMaxLength(300);
+            entity.Property(e => e.Description).HasColumnName("DESCRIPTION").HasMaxLength(2000);
+            entity.HasOne(e => e.ReqState).WithMany().HasForeignKey(e => e.State);
```
iaincrawf, line 844: I'm late with this sorry, but Description is already mapped above as PartDescription - if there's a part then description gets populated with the part.description when making the req so seemed appropriate. State is already mapped as a string too but not sure if you need some other part of the reqState object!

lewisrenfrew: I think the req description is a seperate thing to the PartDescription, so probably makes more sense to stick with Description since it might be different to the Part's description

iaincrawf: The old form had it labelled as part description on the front end but I guess if it's Sundry it's kind of not a part that's being described. Either probably works but whoever of us does a pr next should probably remove the duplicate string mapping to the same field and just keep one
image

PR #155, opened 2022-03-31
react-pdf proof of concept

lewisrenfrew:
nice emailable pdfs with selectable text. multi page support etc.


lewisrenfrew:
nice emailable pdfs with selectable text. multi page support etc.

iaincrawf (approved):

PR #156, opened 2022-03-31
Bluereq ut

iaincrawf:
stayed on same branch so has a longer commit history than it should and still to resolve conflicts

Need to
- add link to new print page
- give finance sign off permission to finance staff 
but this should be it in a useable state aside from the making it into an order bit. Will look at adding notification emails next ie "please look and authorise this req"


src/Domain.LinnApps/PurchaseOrders/PurchaseOrderReqService.cs
```diff
@@ -14,33 +14,38 @@ public class PurchaseOrderReqService : IPurchaseOrderReqService
...
             this.purchaseOrderReqsPack = purchaseOrderReqsPack;
+            this.reqsStateChangeRepository = reqsStateChangeRepository;
         }
 
         public void Authorise(PurchaseOrderReq entity, IEnumerable<string> privileges, int currentUserId)
         {
-            if (!this.authService.HasPermissionFor(AuthorisedAction.PurchaseOrderReqAuthorise, privileges))
+            var stage = !entity.AuthorisedById.HasValue ? "AUTH1" : "AUTH2";
+
+            if (stage == "AUTH1" && entity.State != "AUTHORISE WAIT")
```
lewisrenfrew, line 33: Req state should probably come from the db as opposed to being hardcoded strings

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderReqService.cs
```diff
@@ -100,26 +102,24 @@ public void FinanceApprove(PurchaseOrderReq entity, IEnumerable<string> privileg
...
             {
-                throw new UnauthorisedActionException("Cannot authorise a req that is not in state 'AUTHORISE WAIT'");
+                throw new UnauthorisedActionException("Cannot authorise a req that is not in state 'FINANCE WAIT'. Please make sure the req is saved in this state and try again");
             }
 
-            return;
-            // todo find finance check equivalent of allowed to authorise checks
-            //entity.FinanceCheckById = currentUserId;
+            entity.FinanceCheckById = currentUserId;
+            //todo get next state from state change table instead of hard coding
```
lewisrenfrew, line 109: ah, you already have a todo similar to my comment

lewisrenfrew (approved):

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderReqService.cs
iaincrawf: Like you spotted I do have a todo for when it's changing state to get the next one from the state changes table, but for things like this where I'm checking if it's in the authorise wait state, do you think it's better to do like rough code example below?
`repo.find(x=>x.State == "AUTHORISE WAIT").State`
 It's still kinda hard coded but I guess if we changed the states it'd break the code and then we'd be forced to fix this form, which'd be better I think? There are a few bits like this hard coded in the blue req pack (Kinda need to at some point) and the old form

lewisrenfrew: Yeah, it doesn't really matter too much

PR #157, opened 2022-04-07
Po req emails

iaincrawf:
need to write tests for these new email methods still but I'd like to get more users using it and think this should let them


RossStewart (approved):

RossStewart:
Looks good!

PR #158, opened 2022-04-11
Report exports

lewisrenfrew:
- I didn't really like how we were doing csv responses (all of the csv -> stream business etc repeated across every report module / facade service) so I tried to move that stuff into a reusable method
- The Carter way of doing this sort of thing seems to be Extension methods to the Request/Response objects, so I stuck to that policy
- Let me know what y'all think and if we are happy I'll refactor all the csv exports to work this way
- could add this extension to Common.Carter if we so desire

### DIFF
--- a/src/Service/Modules/Reports/PartsReceivedReportModule.cs
+++ b/src/Service/Modules/Reports/PartsReceivedReportModule.cs
@@ -1,5 +1,4 @@
 ﻿using System.Net.Mime;
-
 namespace Linn.Purchasing.Service.Modules.Reports
 {
     using System.Threading.Tasks;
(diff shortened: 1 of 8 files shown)


src/Service/Modules/Reports/PartsReceivedReportModule.cs
```diff
@@ -55,5 +58,22 @@ private async Task GetReport(HttpRequest req, HttpResponse res)
...
+                    FromDate = req.Query.As<string>("fromDate"),
+                    ToDate = req.Query.As<string>("toDate"),
+                    Jobref = req.Query.As<string>("jobref"),
+                    OrderBy = req.Query.As<string>("orderBy"),
+                    IncludeNegativeValues = req.Query.As<bool>("includeNegativeValues")
+                };
+            
+            var csv = this.reportFacadeService.GetReportCsv(options);
+
+            await res.FromCsv(csv, "parts_received.csv");
```
lewisrenfrew, line 76: all of the csv -> stream business is just replaced with this .FromCsv(..) extension to the HttpResponse

richardiphillips (approved):

PR #159, opened 2022-04-11
Remove permissions to Overbook Search, add it to editing instead

RossStewart:
Allow any user to search for items but only allow users that are able to update Purchase Orders to allow an overbooking.


src/Service.Host/client/src/actions/purchaseOrdersApplicationStateActions.js
```diff
@@ -0,0 +1,13 @@
+import { StateApiActions } from '@linn-it/linn-form-components-library';
+import { purchaseOrdersActionTypes as actionTypes } from './index';
+import * as itemTypes from '../itemTypes';
+import config from '../config';
+
+export default new StateApiActions(
```
lewisrenfrew, line 6: Not 100% sure, but I don't think you need this? I think the purchaseOrdersActions will have a .fetchState() action already

src/Service.Host/client/src/components/OverbooksSearch.js
```diff
@@ -1,15 +1,25 @@
...
+import Tooltip from '@mui/material/Tooltip';
 import purchaseOrdersActions from '../actions/purchaseOrdersActions';
+import purchaseOrdersApplicationStateActions from '../actions/purchaseOrdersApplicationStateActions';
 import history from '../history';
 import config from '../config';
 
 function OverbooksSearch() {
     const dispatch = useDispatch();
+    useEffect(() => dispatch(purchaseOrdersActions.fetch()), [dispatch]);
+    useEffect(() => dispatch(purchaseOrdersApplicationStateActions.fetchState()), [dispatch]);
```
lewisrenfrew, line 22: yeah, I think you can just do purchaseOrdersActions.fetchState() here, and remove line 21 and delete the extraneous  purchaseOrdersApplicationStateActions.js

RossStewart: Ah ok! Cheers for checking that.

Just sorting a GIT issue I'm having atm as my test isn't pushing. Will have that sorted in a sec.

src/Service.Host/client/src/actions/purchaseOrdersApplicationStateActions.js
iaincrawf: Oh yeah so it does, looks like we've done it for reqs and signing limits too. I didn't spot that on FetchApiActions before

src/Service.Host/client/src/components/OverbooksSearch.js
RossStewart: Just a general Q but we have StateAPI actions and FetchAPI actions - what's the relationship here? Is there a benefit to using StateAPI actions over FetchAPI in select circumstances? Will have a look at the common code but just wanted to check in first.

richardiphillips: only if you have no need for fetch api. they started separate but got added to fetch and update for convenience
 from memory

lewisrenfrew (approved):

PR #162, opened 2022-04-14
Reciptpsdiff

andrewfraser73:
Simple report added to compare receipts vs prefsup price and display difference.  Will probably end up being drill down from dashboard

Doesn't have a snappy acronym like MPV or CPV suggestions welcome.


src/Domain.LinnApps/Reports/Models/ReceiptPrefsupDiff.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace Linn.Purchasing.Domain.LinnApps.Reports.Models
+{
+    using System;
+
+    public class ReceiptPrefSupDiff
+    {
+        public int PlReceiptId { get; set; }
+
+
```
lewisrenfrew, line 9: blankity blank

lewisrenfrew (approved):

src/Domain.LinnApps/Reports/IPrefSupReceiptsReportService.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    using System;
+
+    using Linn.Common.Reporting.Models;
+
+    public interface IPrefSupReceiptsReportService
+    {
+        ResultsModel GetReport(DateTime fromDate, DateTime toDate);
+    }
+}
```
iaincrawf, line 11: there's quite a few of these red sign thingies for missing end of file line breaks, not sure if we can get stylecop to complain about them in visual studio? I think I've managed to get a code format command to add them on mine if I remember to run that

src/Persistence.LinnApps/Repositories/ReceiptPrefSupDiffRepository.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        public ReceiptPrefSupDiff FindBy(Expression<Func<ReceiptPrefSupDiff, bool>> expression)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IQueryable<ReceiptPrefSupDiff> FilterBy(Expression<Func<ReceiptPrefSupDiff, bool>> expression)
+        {
+            return this.serviceDbContext.ReceiptPrefsupDiffs.Where(expression);
```
iaincrawf, line 26: I think since there aren't any includes or custom code this repository file could go and could just do the sort of new EntityFrameworkQueryRepository in the IOC PersistenceExtensions class thing. Probably doesn't matter that it's there but it might save you making an extra class in future https://github.com/linn/common/blob/master/src/Linn.Common.Persistence.EntityFramework/EntityFrameworkQueryRepository.cs

PR #165, opened 2022-04-20
carter 6

richardiphillips:


src/Service/Modules/Reports/PartsReceivedReportModule.cs
```diff
@@ -13,66 +12,81 @@
...
-        private async Task GetReport(HttpRequest req, HttpResponse res)
+        private async Task GetReport(
+            HttpRequest req,
+            HttpResponse res,
+            int? supplier,
+            string fromDate,
+            string toDate,
+            string jobref,
+            string orderBy,
+            bool includeNegativeValues,
```
lewisrenfrew, line 45: this is way nicer than the rq.Query syntax

richardiphillips: yeah. i agree

iaincrawf (approved):
Clicked about on sys and all seems ok to me. Good shout separating reqs and orders modules too, been meaning to do that

src/Service/Modules/Reports/PartsReceivedReportModule.cs
iaincrawf: What do you guys think of doing this to make bigger resource objects too? I hate that when something is null or js has passed something back as a string not an int etc it just fails completely silently with the bind<resource> way, wondering if this would avoid that. It would look pretty horrible with 20 parameters or something though so maybe not

src/Service/Modules/Reports/PartsReceivedReportModule.cs
```diff
@@ -13,66 +12,81 @@
...
-                                  OrderBy = req.Query.As<string>("orderBy"),
-                                  IncludeNegativeValues = req.Query.As<bool>("includeNegativeValues")
+                                  Supplier = supplier,
+                                  FromDate = fromDate,
+                                  ToDate = toDate, 
+                                  Jobref = jobref, 
+                                  OrderBy = orderBy, 
+                                  IncludeNegativeValues = includeNegativeValues
                               };
             if (string.IsNullOrEmpty(options.ToDate))
```
lewisrenfrew, line 57: I'm using the presence or absence of a required option here to decide whether to return the report or just the view... Could that be causing a problem with the new syntax?

lewisrenfrew: Seems not, since I do the same for https://app-sys.linn.co.uk/purchasing/reports/whats-due-in which looks ok

PR #168, opened 2022-04-21
Whats insp report v2

lewisrenfrew:


src/Resources/WhatsInInspectionReportResource.cs
```diff
@@ -1,19 +1,13 @@
...
-
-        public string OurUnitOfMeasure { get; set; }
-
-        public decimal QtyInStock { get; set; }
-
-        public decimal QtyInInspection { get; set; }
+        public IEnumerable<WhatsInInspectionReportEntryResource> PartsInInspection { get; set; }
 
-        public ReportReturnResource OrdersBreakdown { get; set; }
+        public ReportReturnResource BackOrderData { get; set; }
```
richardiphillips, line 11: are they really back orders?

richardiphillips (approved):
phew complicated

src/Resources/WhatsInInspectionReportResource.cs
lewisrenfrew: I'm not sure really. This little ResultsModel/ReportReturnResource is just a copy of the data in this footer of the old report
Screenshot 2022-04-21 092248

Do you think there's a better name for it?

richardiphillips: no fair enough. perfect name

PR #171, opened 2022-04-25
Receipts feedback

andrewfraser73:
a couple of bits of Andrew McV feedback on the report and the ability to call it with query strings from the dashboard


src/Service.Host/client/src/components/reports/PrefSupReceiptsReport.js
```diff
@@ -17,10 +18,11 @@ import prefSupReceiptsReportActions from '../../actions/prefSupReceiptsReportAct
 function PrefSupReceiptsReport() {
     const dispatch = useDispatch();
 
+    const queryOptions = queryString.parse(window.location.search);
```
richardiphillips, line 21: trying to get rid of the window._ calls.   useLocation() is much better.

richardiphillips (approved):

PR #174, opened 2022-04-26
sorting on report

andrewfraser73:
simple one file report to sort report how andrew wants it sorted

### DIFF
--- a/src/Domain.LinnApps/Reports/PrefSupReceiptsReportService.cs
+++ b/src/Domain.LinnApps/Reports/PrefSupReceiptsReportService.cs
@@ -8,6 +8,8 @@
     using Linn.Common.Reporting.Models;
     using Linn.Purchasing.Domain.LinnApps.Reports.Models;
 
+    using MoreLinq;
+
     public class PrefSupReceiptsReportService : IPrefSupReceiptsReportService
     {
         private readonly IQueryRepository<ReceiptPrefSupDiff> receiptRepository;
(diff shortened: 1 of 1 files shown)


src/Domain.LinnApps/Reports/PrefSupReceiptsReportService.cs
```diff
@@ -59,7 +61,9 @@ public string GetReportCurrencyValue(decimal currencyValue, string currency, dec
         public ResultsModel GetReport(DateTime fromDate, DateTime toDate)
         {
             var results = this.receiptRepository.FilterBy(
-                e => e.DateBooked >= fromDate && e.DateBooked <= toDate && e.Difference != 0);
+                e => e.DateBooked >= fromDate && e.DateBooked <= toDate && e.Difference != 0).OrderBy(
+                e => e.Difference,
+                OrderByDirection.Descending);
```
richardiphillips, line 66: is this faster than sorting outwith the repo? Does morelinq make the native db calls ?

richardiphillips (approved):

src/Domain.LinnApps/Reports/PrefSupReceiptsReportService.cs
andrewfraser73: didn't need any speed difference but was hoping it would make its way into SQL (could alternatively stick it in to the View definition but that hides it a bit).

richardiphillips: if it's in the view then at least it's liable for database optimisation. if it's in the domain service then it's testable domain logic.  If morelinq doesn't make native code then it's untestable code and reliant for a third party library for no gain?

PR #178, opened 2022-04-27
reqs - cc sender in authorisation request emails

iaincrawf:
+ don't change req number - was effectively deleting them if you updated anything

### DIFF
--- a/src/Domain.LinnApps/PurchaseOrderReqs/PurchaseOrderReqService.cs
+++ b/src/Domain.LinnApps/PurchaseOrderReqs/PurchaseOrderReqService.cs
@@ -252,11 +252,6 @@
             entity.QuoteRef = updatedEntity.QuoteRef;
             entity.Email = updatedEntity.Email;
             entity.DateRequired = updatedEntity.DateRequired;
-            entity.RequestedBy = updatedEntity.RequestedBy;
-            entity.AuthorisedBy = updatedEntity.AuthorisedBy;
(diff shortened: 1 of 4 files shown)


richardiphillips (approved):

richardiphillips:
I noticed you had hardcoded urls in the domain service e.g. https://app.linn.co.uk/purchasing..

Would be much better to pass the root in as a variable or look it up. I'd prefer to get ioc to look up APP_ROOT and pass it in

iaincrawf:
> I noticed you had hardcoded urls in the domain service e.g. https://app.linn.co.uk/purchasing..
> 
> Would be much better to pass the root in as a variable or look it up. I'd prefer to get ioc to look up APP_ROOT and pass it in

I remember thinking I should change that at the time and forgot, is using ConfigurationManager["APP_ROOT"] what you mean with IOC or something else? It seems to work ok

richardiphillips:
> 

having the configuration manager fishing things out isn't really nice domain code. Much better to make it a constructor variable and pass it in from the IOC. There should be examples in the Messaging Extensions.

richardiphillips:
> > 
> 
> having the configuration manager fishing things out isn't really nice domain code. Much better to make it a constructor variable and pass it in from the IOC. There should be examples in the Messaging Extensions.

Actually see how Lewis did the PdfService in ServiceExtensions. That means you can then test that the right uris are being generated too

PR #181, opened 2022-04-28
remove damages % field from client as per purchasing ticket

richardiphillips:


lewisrenfrew:
iirc the domain service complains if you try and save without a damages percent (since it was a required field on oracle)

PR #188, opened 2022-05-04
Mrp Run

richardiphillips:


src/Service.Host/client/src/components/__tests__/RunMrp.test.js
```diff
@@ -0,0 +1,271 @@
...
+        loading: false
+    }
+};
+const stateWithMrpRunResult = {
+    ...stateWithMasterNotRunning,
+    runMrp: {
+        data: {
+            success: true,
+            message: 'MRP requested with run log id 11663',
+            processHref: '/purchasing/material-requirements/run-logs/11663'
```
lewisrenfrew, line 94: Should have said this on the last PR, but i thought putting the processHref in the links array of the resource was more in line with all our hateoas stuff

lewisrenfrew (approved):

src/Service.Host/client/src/components/__tests__/RunMrp.test.js
richardiphillips: it's a very fair point. especially when you mentioned the possibility of a cancel href as per mailchimp

PR #196, opened 2022-05-11
Acknowledge orders screen

lewisrenfrew:
- I decided to settle on a slightly different (and hopefully more flexible? if not quite as good as full on Visitor Pattern goodness) way of enacting a PATCH in my FacadeService
- Yesterday like I showed I was just locking the Patch facade method down to the very specific type of patch I wanted to do from my form which wasn't ideal since a future PATCH-er of a PO delivery would need to add a whole new PATCH method or something 
- my new idea is basically to have a <code>PatchResource< T></code>, with a <code>T from </code> and a <code>T To</code> field. The facade service can then look at these objects and update whatever it needs to update .
- This way the client can decide what it wants to patch by providing a diff in the form of two objects and the server can handle any possible diff in one Patch facade method. Can show and tell if that's unclear but until then... Any thoughts?
- Could easily add all this to common if we so decide. Would just need to add an 
<code>abstract IResult< TResource> Patch(PatchRequestResource< T> resource)</code> method to the common FacadeService


Purchasing.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.31702.278
+# Visual Studio Version 17
```
iaincrawf, line 3: did you deliberately upgrade this or did vs do it on its own?

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public PurchaseOrderDeliveryService(IRepository<PurchaseOrderDelivery, PurchaseOrderDeliveryKey> repository)
+        {
+            this.repository = repository;
+        }
+
+        public IEnumerable<PurchaseOrderDelivery> SearchDeliveries(
+            string supplierSearchTerm, string orderNumberSearchTerm, bool includeAcknowledged)
+        {
+            var result = this.repository.FindAll();
```
iaincrawf, line 21: findAll initially seemed very bold but because you're not .toList ing etc is it just when it gets to the order by it'll evaluate, and by that time have some wheres on it to cut it down?

src/Facade/Services/PurchaseOrderDeliveryFacadeService.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        public IResult<PurchaseOrderDeliveryResource> PatchDelivery(
+            PurchaseOrderDeliveryKey key,
+            PatchRequestResource<PurchaseOrderDeliveryResource> requestResource, 
+            IEnumerable<string> privileges)
+        {
+            var entity = this.repository.FindById(key);
+
+            if (requestResource.From.DateAdvised != requestResource.To.DateAdvised)
```
iaincrawf, line 55: would it be safer for these to compare entity.DateAdvised etc to the resource in this method?

iaincrawf (approved):
Looks good

Purchasing.sln
lewisrenfrew: did it on its own

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
lewisrenfrew: yeah.. its just so I can build up the query in the domain service and test it since there's a bit of logic involved which would otherwise just get hidden away in the FilterBy(expression)

src/Facade/Services/PurchaseOrderDeliveryFacadeService.cs
lewisrenfrew: That's a good question. 

In my head the person requesting the PATCH defines a 'from' and a 'to' state and that's what's used to decide whether to apply an update to any given field or not. 

So I think comparing the two things in the resource is probably correct.. Feel free to disagree though

iaincrawf: Yeah fair enough, I was just thinking what if we accidentally altered the 'from' state before sending it to the backend here, but actually I can't really see a scenario where it matters.. Unless it happened to be the same as the new To and so didn't update maybe?

iaincrawf: But yeah either using part of the resource or the entity as the from state will be fine I'm sure, and if both in resource seems more pure then go for it

lewisrenfrew: I'll earmark this for the next show & tell and we can discuss it before I think about adding it to the Common code

PR #198, opened 2022-05-12
Acknowledge orders v2

lewisrenfrew:
- file uploader
- far from complete but jacki is off till the 17th so just parking in this minimum working state


src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
```diff
@@ -44,5 +53,100 @@ public PurchaseOrderDeliveryService(IRepository<PurchaseOrderDelivery, PurchaseO
...
+            PurchaseOrderDelivery from,
+            PurchaseOrderDelivery to,
+            IEnumerable<string> privileges)
+        {
+            if (!this.authService.HasPermissionFor(AuthorisedAction.PurchaseOrderUpdate, privileges))
+            {
+                throw new UnauthorisedActionException("You are not authorised to acknowledge orders.");
+            }
+
+            var entity = this.repository.FindById(key);
```
iaincrawf, line 68: I think you've already done this findById in the facade service and actually update these fields there too after this - do you want to pass the entity in from there and update them by reference here sort of thing? Probs at least stop duplicate these field them unless I've read it wrong

tests/Integration/Integration.Tests/PurchaseOrderDeliveryModuleTests/WhenBatchUpdatingAndSucces.cs
```diff
@@ -0,0 +1,66 @@
...
+    using Linn.Purchasing.Domain.LinnApps;
+    using Linn.Purchasing.Domain.LinnApps.PurchaseOrders;
+    using Linn.Purchasing.Integration.Tests.Extensions;
+    using Linn.Purchasing.Resources;
+
+    using NSubstitute;
+
+    using NUnit.Framework;
+
+    public class WhenBatchUpdatingAndSucces : ContextBase
```
iaincrawf, line 17: doesn't matter but missing an 's' in success 😄

iaincrawf (approved):

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
lewisrenfrew: I just forgot to delete the field updates from the facade when I refactored the update into the domain. Good spot!

iaincrawf: Ah yeah makes sense. There's still the 2 findById calls - do you wanna pass the entity from the facade to the domain or happy it'll be updated by reference/ef magic by the time it gets send to the resource builder?

lewisrenfrew: No, I can completely remove any persistence dependencies from my facade now since it's all done in the domain

PR #206, opened 2022-05-18
Edi

andrewfraser73:
First draft of EDI program

- it sends Emails via Oracle because the PL/SQL package makes a text file on the Oracle W: drive and to rewrite that is a bigger job.

- it links to the new purchase order viewer screen although they might not use that functionality much

- it requires new PL/SQL package EDI_EMAIL_PACK that does not much on DEV but will send it for real on LINN


src/Service.Host/client/src/components/EdiOrders.js
```diff
@@ -0,0 +1,224 @@
...
+                        Get orders
+                    </Button>
+                </Grid>
+                <Grid item xs={9}>
+                    {emailSending ? (
+                        <Loading />
+                    ) : (
+                        <Button
+                            variant="contained"
+                            disabled={!ediOrders || !ediOrders.length || emailSentResult}
```
richardiphillips, line 203: are there tests for this component?

PR #207, opened 2022-05-19
Split deliveries etc

lewisrenfrew:
- just the backend since I deliberated over it for so long


tests/Unit/Domain.LinnApps.Tests/Domain.LinnApps.Tests.csproj
```diff
@@ -23,7 +23,6 @@
   <ItemGroup>
     <Folder Include="ForecastingServiceTests\" />
     <Folder Include="MrUsedOnReportServiceTests\" />
-    <Folder Include="PurchaseOrderDeliveryServiceTests\" />
```
lewisrenfrew, line 26: huh

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDelivery.cs
```diff
@@ -6,7 +6,7 @@ public class PurchaseOrderDelivery
     {
         public string Cancelled { get; set; }
 
-        public DateTime? DateAdvised { get; set; }
+        public DateTime? AdvisedDate { get; set; }
```
iaincrawf, line 9: I kinda prefer dateAdvised for consistency with other DateBlahs but not a biggie

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -96,42 +97,11 @@ public PurchaseOrder UpdateOrder(PurchaseOrder current, PurchaseOrder updated, I
...
         }
 
-        private void UpdateDetails(ICollection<PurchaseOrderDetail> currentDetails, ICollection<PurchaseOrderDetail> updatedDetails)
-        {
-            foreach (var detail in updatedDetails)
-            {
-                var currentDetail = currentDetails.FirstOrDefault(x => x.Line == detail.Line);
-                if (currentDetail != null)
-                {
-                    this.UpdateDetailProperties(currentDetail, detail);
```
iaincrawf, line 112: I'll still need most of this method to update the purchase order details from the po utility, just not foreach with deliveries and the updateDeliveryProperties method

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
```diff
@@ -77,11 +90,21 @@ public class PurchaseOrderDeliveryService : IPurchaseOrderDeliveryService
...
+                throw new UnauthorisedActionException("Orders are currently restricted.");
+            }
+
             var entity = this.repository.FindById(key);
+            var miniOrder = this.miniOrderRepository.FindById(key.OrderNumber);
+            var miniOrderDelivery = this.miniOrderDeliveryRepository.FindBy(
+                x => x.OrderNumber == key.OrderNumber && x.DeliverySequence == key.DeliverySequence);
 
-            if (from.DateAdvised != to.DateAdvised)
+            if (from.AdvisedDate != to.AdvisedDate)
```
iaincrawf, line 103: I still feel funny about checking something other than the entity pulled from the db, in order to determine whether to update the entity from the db but we've already had a pr discussion about that 😄

iaincrawf (approved):
cool, just a wee bit of the po service to go back in I think. The mini order stuff looks easy enough to remove once we get a real po form allowing multiline purchase orders which is good

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
iaincrawf: (It wasn't referenced even in the interface yet but I will need to use it eventually, or a completed/tested version of it anyway)

src/Facade/Services/PurchaseOrderDeliveryFacadeService.cs
```diff
@@ -135,11 +135,66 @@ public class PurchaseOrderDeliveryFacadeService : IPurchaseOrderDeliveryFacadeSe
...
+            int orderNumber, 
+            int orderLine,
+            IEnumerable<PurchaseOrderDeliveryResource> resource,
+            IEnumerable<string> privileges)
+        {
+            var purchaseOrderDeliveryResources = resource.ToList();
+            this.domainService.UpdateDeliveriesForOrderLine(
+                orderNumber,
+                orderLine,
+                purchaseOrderDeliveryResources.Select(d => new PurchaseOrderDelivery
```
iaincrawf, line 148: Could this use the buildEntityFromResource method you've got or is it different?

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
lewisrenfrew: Yeah - will go over at show & tell and we can discuss

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
lewisrenfrew: I think I've accidentally removed this - only meant to remove the delivery part

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDelivery.cs
lewisrenfrew: I can't recall why I changed it.. Possibly because that's what it's called on the db table

iaincrawf: Yeah I'm pretty sure it is called that on the table, maybe it was for consistency with mini Order Delivery etc if you'd called them AdvisedDate? I don't really mind either way as we know what it is both ways round

PR #210, opened 2022-05-20
Shortages Report

RossStewart:
Drilldowns will be in subsequent PRs, will be adding JS tests too. Submitting to get a general consensus.


RossStewart:
Will fix conflicts on Monday morn

src/Domain.LinnApps/Reports/ShortagesReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            this.shortagesEntryView = shortagesEntryView;
+        }
+
+        public IEnumerable<ResultsModel> GetReport(
+            int purchaseLevel,
+            string vendorManager)
+        {
+            
+            var results = this.shortagesEntryView.FilterBy(x => (Convert.ToInt32(x.PurchaseLevel) <= purchaseLevel) &&
```
lewisrenfrew, line 26: do we need this Convert.ToInt32? (Not saying we don't - I just don't think I've had to do this before)

src/Domain.LinnApps/Reports/ShortagesReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            foreach (var shortagesForPlanner in results.GroupBy(a => new { a.PlannerName }))
+            {
+                var model = new ResultsModel();
+
+                model.AddColumn("VendorManagerCode", "Vendor Manager Code");
+                model.AddColumn("VendorManagerName", "Vendor Manager Name");
+
+                model.ReportTitle = new NameModel(
+                    $"Purchasing shortages planner");
```
lewisrenfrew, line 39: You should include the Planners name in each ResultsModels title

src/Domain.LinnApps/Reports/ShortagesReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+                model.RowHeader = "Planner";
+
+                var distinctVendorManagers = shortagesForPlanner.DistinctBy(x => x.VendorManagerCode);
+                var distinctLevels = shortagesForPlanner.DistinctBy(x => x.PurchaseLevel);
+
+                foreach (var level in distinctLevels)
```
lewisrenfrew, line 45: you should probably order this list before looping throught it and adding columns (Otherwise your columns could be in a random order?

src/Facade/Services/ShortagesReportFacadeService.cs
```diff
@@ -0,0 +1,39 @@
...
+            IShortagesReportService domainService,
+            IBuilder<ResultsModel> resultsModelResourceBuilder)
+        {
+            this.domainService = domainService;
+            this.resultsModelResourceBuilder = resultsModelResourceBuilder;
+        }
+
+        public IResult<IEnumerable<ReportReturnResource>> GetReport(ShortagesReportRequestResource options)
+        {
+            var purchaseLevel = int.Parse(options.PurchaseLevel);
```
lewisrenfrew, line 28: I would make options.PurchaseLevel an int in the resource so you don't have to parse it.

src/Domain.LinnApps/Reports/ShortagesReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Linn.Common.Persistence;
+    using Linn.Common.Reporting.Models;
+    using Linn.Purchasing.Domain.LinnApps.Reports.Models;
+
+    public class ShortagesReportService : IShortagesReportService
+    {
+        private readonly IRepository<ShortagesEntry, string> shortagesEntryView;
```
lewisrenfrew, line 13: Since this is mapped to a View (not a table) and you've declared it as having .HasNoKey() in the ServiceDbContext, you shouldn't really use IRepopsitory (since you're implying a key type of string here when there is no key). IQueryRepository lets you define a repository without a key type which is more suitable

src/Service.Host/client/src/components/reports/ShortagesReport.js
```diff
@@ -0,0 +1,87 @@
...
+
+    return (
+        <>
+            <Grid style={{ marginTop: 40 }} container spacing={3} justifyContent="center">
+                <Grid item xs={12}>
+                    <BackButton backClick={() => handleBackClick(history, options)} />
+                </Grid>
+                <Grid item xs={12}>
+                    {!loading && reportData ? (
+                        <ExportButton
```
lewisrenfrew, line 47: You haven't written an export route so this button won't work? I'm not sure this report needs an export anyways so I'd just delete this buton.

src/Service.Host/client/src/components/reports/ShortagesReport.js
```diff
@@ -0,0 +1,87 @@
...
+                        <></>
+                    )}
+                </Grid>
+                <Grid item xs={12}>
+                    {loading || !reportData ? (
+                        <Loading />
+                    ) : (
+                        <>
+                            {reportData.map(report => (
+                                <ReportTable
```
lewisrenfrew, line 64: elements resulting from a list.map(..) should have keys to stop react complaining:
https://reactjs.org/docs/lists-and-keys.html#keys

Also did the MutliReporttable component not work here?

tests/Unit/Domain.LinnApps.Tests/ShortagesReportTests/ContextBase.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        protected IShortagesReportService Sut { get; private set; }
+
+        protected IRepository<ShortagesEntry, string> ShortagesRepository { get; private set; }
+
+        protected IEnumerable<ShortagesEntry> Data { get; private set; }
+
+        protected IReportingHelper ReportingHelper { get; private set; }
+
+
```
lewisrenfrew, line 27: blank

tests/Unit/Domain.LinnApps.Tests/ShortagesReportTests/WhenGettingReport.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        private IEnumerable<ResultsModel> results;
+
+        [SetUp]
+        public void SetUp()
+        {
+            this.results = this.Sut.GetReport(3, "ALL");
+        }
+
+
```
lewisrenfrew, line 27: blank

tests/Unit/Domain.LinnApps.Tests/ShortagesReportTests/WhenGettingReport.cs
```diff
@@ -0,0 +1,41 @@
...
+        [Test]
+        public void ShouldGetData()
+        {
+            this.ShortagesRepository.Received().FilterBy(Arg.Any<Expression<Func<ShortagesEntry, bool>>>());
+        }
+
+        [Test]
+        public void ShouldReturnData()
+        {
+            this.results.Should().HaveCount(2);
```
lewisrenfrew, line 37: Since your report domain service is doing some grouping and totalling etc. I think you should test the correct results are actually built here (examples of checking report data here https://github.com/linn/purchasing/blob/main/tests/Unit/Domain.LinnApps.Tests/PurchaseOrdersReportServiceTests/WhenGettingOrdsBySupplierReportStockN.cs

You are also filtering your data in your domain service so you should have tests for different options scenarios e.g.
WhenPurchaseLevel3 or WhenVendorManagerL

lewisrenfrew (approved):

src/IoC/PersistenceExtensions.cs
```diff
@@ -110,7 +110,9 @@ public static IServiceCollection AddPersistence(this IServiceCollection services
                 .AddTransient<IRepository<LedgerPeriod, int>, EntityFrameworkRepository<LedgerPeriod, int>>(r
                     => new EntityFrameworkRepository<LedgerPeriod, int>(r.GetService<ServiceDbContext>()?.LedgerPeriods))
                 .AddTransient<IRepository<CancelledOrderDetail, int>, EntityFrameworkRepository<CancelledOrderDetail, int>>(
-                    r => new EntityFrameworkRepository<CancelledOrderDetail, int>(r.GetService<ServiceDbContext>()?.CancelledPurchaseOrderDetails));
+                    r => new EntityFrameworkRepository<CancelledOrderDetail, int>(r.GetService<ServiceDbContext>()?.CancelledPurchaseOrderDetails))
+                .AddTransient<IRepository<ShortagesEntry, string>, EntityFrameworkRepository<ShortagesEntry, string>>(r
```
lewisrenfrew, line 114: as per previous comment, Should really be an IQueryRepository<ShortagesEntry>

lewisrenfrew (approved):
Look good. Could do with more / better unit tests

src/Domain.LinnApps/Reports/ShortagesReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.shortagesEntryView = shortagesEntryView;
+        }
+
+        public IEnumerable<ResultsModel> GetReport(
+            int purchaseLevel,
+            string vendorManager)
+        {
+            
+            var results = this.shortagesEntryView.FilterBy(x => (Convert.ToInt32(x.PurchaseLevel) <= purchaseLevel) &&
+                                                                            (vendorManager == "ALL" || vendorManager == x.VendorManagerCode)).ToList();
```
lewisrenfrew, line 27: how do you actually run this report for vendorManager = "ALL" ? I can't see how you could select 'ALL' from your options screen dropdown

src/Facade/Services/ShortagesReportFacadeService.cs
RossStewart: This was a thing that was tripping me a little as the table has it marked as a string. As far as I can tell it will always be an integer. Would cause the query to blow up as it would say that it could not convert from a string to an int. Happy to change it but how could I get around this at the query level?

src/Domain.LinnApps/Reports/ShortagesReportService.cs
RossStewart: OK, noted!

src/IoC/PersistenceExtensions.cs
RossStewart: Noted

src/Domain.LinnApps/Reports/ShortagesReportService.cs
RossStewart: This was something I wasn't entirely sure on. I had tried to Parse/TryParse this value but LINQ just wasn't having it - only ConvertToInt32 was valid. If I can change the field to an int I could do away with it.

src/Domain.LinnApps/Reports/ShortagesReportService.cs
RossStewart: Will add in.

src/Domain.LinnApps/Reports/ShortagesReportService.cs
RossStewart: Was going to ask about this. Will add in.

src/Domain.LinnApps/Reports/ShortagesReportService.cs
lewisrenfrew: Ah I see... It really feels like the PURCH_LEVEL column should be a number and not a varchar. Andrew or Richard might know a reason why it isn't? It's PURCH_LEVEL on V_SHORT_TRIGGER_LEVELS.

you could force it to a number in your view definition? Or just leave it a string. If you leave it a string this code is fine

richardiphillips: I don't know why it's a string but it's definitely set as a string in the view so I wouldn't ever not present it as a string

lewisrenfrew: But you should still change it to be an int in your resource and drop the parse in your facade service.
right now it goes from:
an int from your options form -> to a string in your resource -> parsed back to an int in your facade -> passed as an int to your domain  
which is lots of pointless conversions

lewisrenfrew: Should just be an int all the way imo, and then this line 26 here will deal with the database type conversions (if you don't change your view to make the column a number). Hope that makes sense

lewisrenfrew: @richardiphillips we wrote the view for this report... could we just convert it to a number in the view sql? It feels like it really should be a number to me but I agree the C# should reflect the view properly

richardiphillips: ah ok. yeah i agree if you wrote the view for this report then just make it a number

PR #211, opened 2022-05-24
po reqs - fix decimal places being cut off

iaincrawf:
was parsing int so total never had decimal places and so some are wrong. Will check recents

### DIFF
--- a/src/Service.Host/client/src/components/POReqs/POReqUtility.js
+++ b/src/Service.Host/client/src/components/POReqs/POReqUtility.js
@@ -365,11 +365,7 @@ function POReqUtility({ creating }) {
     const [alreadyShownCostWarning, setAlreadyShownCostWarning] = useState(false);
 
     useEffect(() => {
-        if (
-            req.unitPrice &&
-            req.qty &&
-            (creating || (req.unitPrice !== item?.unitPrice && req.qty !== item?.qty))
-        ) {
+        if (req.unitPrice && req.qty) {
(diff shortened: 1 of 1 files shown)


src/Service.Host/client/src/components/POReqs/POReqUtility.js
```diff
@@ -844,7 +832,7 @@ function POReqUtility({ creating }) {
                             <Grid item xs={4}>
                                 <InputField
                                     fullWidth
-                                    value={req.totalReqPrice}
+                                    value={parseFloat(req.totalReqPrice).toFixed(2)}
```
richardiphillips, line 835: why do we have to parse to a float from a number field?

iaincrawf: was just using the toFixed(2) to force it to always show 2 decimal places even when pulled straight from the db - Steph asked if it could display 2 decimal places for this, carriage and unit price so sort of did this one as a compromise (to fixed on the others messes up typing but this one is disabled and calculated from others). Maybe better leaving them consistent and removing this again? She also wanted commas for big numbers but think going to leave that for now unless she asks again

iaincrawf: I might put it on my list to look at inputField component to try add a comma formatting option and fixed no of decimal places for numbers, but doesn't seem the most important thing right now

lewisrenfrew (approved):

PR #218, opened 2022-06-08
Orders reports fix + deliveries fix

iaincrawf:


src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDelivery.cs
```diff
@@ -57,11 +57,7 @@ public class PurchaseOrderDelivery
         public string CallOffRef { get; set; }
 
         public string FilCancelled { get; set; }
-        
-        public decimal? QtyPassedForPayment { get; set; }
```
richardiphillips, line 61: you definitely need these fields on deliveries. 

Some orders are call off orders (or kanban or JIT) which means we agree to buy 1000 say and just ask for them as we need them. 

This means that sometimes we agree a price (on the order detail) but when we eventually take the items the price may change ( on the delivery)

Hence why the order detail price is call next..price (it's never the definitive price just an expected price)

lewisrenfrew (approved):

PR #220, opened 2022-06-08
don't include values for second delivery in report totals

iaincrawf:
+ tests. Just not displaying the values that get totaled up after the first delivery. Funny wee bool instead of checking seq == 1 is me being paranoid that we delete the first delivery, not sure whether that's possible or not but don't think it's a big deal

### DIFF
--- a/src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
+++ b/src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
@@ -85,6 +85,8 @@
 
                 foreach (var orderDetail in order.Details.Where(d => d.Part.PartNumber == partNumber))
                 {
+                    var isFirstDelivery = true;
+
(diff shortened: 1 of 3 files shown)


src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -386,7 +391,8 @@ private static void AddPartReportColumns(SimpleGridLayout reportLayout)
             PurchaseOrderDetail orderDetail,
             PurchaseOrderDelivery delivery,
             decimal ledgerQty,
-            string currencyCode)
+            string currencyCode,
+            bool isFirstDelivery)
```
lewisrenfrew, line 395: do you need this isFirstDelivery bool? Doesn't the delivery itself tell you whether its the first delivery or not?

lewisrenfrew (approved):

PR #222, opened 2022-06-09
Tickets 15815 15803

lewisrenfrew:


src/Service.Host/client/src/helpers/handleBackClick.js
```diff
@@ -0,0 +1,12 @@
+export default (previousPaths, goBack) => {
+    const prevPath =
+        previousPaths?.[previousPaths.length - 1]?.path +
+        previousPaths?.[previousPaths.length - 1]?.search;
+    if (prevPath.includes('signin-oidc')) {
+        global.window.history.go(-3);
```
iaincrawf, line 6: madness, pretty cool though will be nice if this solves back-button pains

iaincrawf (approved):
lgtm

src/Service.Host/client/src/helpers/handleBackClick.js
iaincrawf: although worth testing if you go to a page and do something like create an order or authorise where it reloads whether it still goes through the previous un-authorised state before actually going back.. either way an improvement over just going to purchasing all the time

lewisrenfrew: Not too sure what you mean. As long as neither of those things redirect then this shouldn't be affected. Do either of those things reidrect (do a history.push() or similar?)

iaincrawf: same as changing a value and saving I think, just sort of reloads. Alison had said when she clicked the browser back it took her to the same req but unauthorised again, which to me was like it was in the browser history with the old state magically saved

PR #223, opened 2022-06-09
Mr tags

richardiphillips:


src/Resources/MaterialRequirements/MrDetailResource.cs
```diff
@@ -37,5 +37,31 @@ public class MrDetailResource
         public string Week12 { get; set; }
         
         public MrDetailItemResource ImmediateItem { get; set; }
+
```
andrewfraser73, line 40: why are there only 12 weeks - is that just for during development and they'll be more later or some other reason.

lewisrenfrew (approved):

andrewfraser73 (approved):
seems alright although not familiar with mr tags to know how it should work

src/Resources/MaterialRequirements/MrDetailResource.cs
richardiphillips: each block shows relative weeks 0 -12 so it's a quarter basically. They can scroll through quarter by quarter or show them all at once as they wish.

PR #230, opened 2022-06-15
fix dates

lewisrenfrew:


src/Facade/Services/PurchaseOrderDeliveryFacadeService.cs
```diff
@@ -86,10 +86,10 @@ public class PurchaseOrderDeliveryFacadeService : IPurchaseOrderDeliveryFacadeSe
 
                     var firstFormatSatisfied =
                         DateTime.TryParseExact(row[2]
-                            .Trim(), "dd/MM/yyyy", CultureInfo.InvariantCulture, DateTimeStyles.None, out var parsedDate);
+                            .Trim(), "dd'/'M'/'yyyy", CultureInfo.InvariantCulture, DateTimeStyles.None, out var parsedDate1);
```
iaincrawf, line 89: Do you need the single quotes inside? Also what's one M, full month?

iaincrawf (approved):

src/Facade/Services/PurchaseOrderDeliveryFacadeService.cs
lewisrenfrew: I'm not sure really. It didn't work so i google it and this suggestion worked

PR #232, opened 2022-06-17
Mr order book Report

lewisrenfrew:
- not quite finished but PR'ing now cos this could get ugly with merge conflicts...
- rewrote the ResultsModelResourceBuilder (and renamed it to ReportReturnResourceBuilder cos thats what it really is) in common. It now handles Lists of ResultsModels as well as a sole ResultsModel. ~~You can make it handle a single ResultsModel by just passing it a List containing a single ResultsModel~~ 
- It doesn't take a claims list to build the Links collection - you'll need to just do this yourself in the facade if necessary... although I don't anticipate we will often need claims when building ReportReturnResources
- I noticed a few places where a claims list was being passed needlessly to report facade services (and then seemingly not used?). Ripped them all out, but double check and let me know if you disagree


src/Facade/Services/IPurchaseOrderReportFacadeService.cs
```diff
@@ -11,31 +11,24 @@
     public interface IPurchaseOrderReportFacadeService
     {
         IResult<ReportReturnResource> GetOrdersBySupplierReport(
-            OrdersBySupplierSearchResource resource,
-            IEnumerable<string> privileges);
+            OrdersBySupplierSearchResource resource);
 
         IEnumerable<IEnumerable<string>> GetOrdersBySupplierExport(
-            OrdersBySupplierSearchResource resource,
-            IEnumerable<string> privileges);
```
lewisrenfrew, line 19: was there any need for all of these methods to tkae a privileges list?

richardiphillips: probably nothing in purchasing really needs it so far. There will definitely be finance reports etc that are restricted. Don't know if they will ever need the privileges passed through to the facade though

richardiphillips:
> * not quite finished but PR'ing now cos this could get ugly with merge conflicts...
> * rewrote the ResultsModelResourceBuilder (and renamed it to ReportReturnResourceBuilder cos thats what it really is) in common. It now handles Lists of ResultsModels as opposed to a sole ResultsModel. You can make it handle a single ResultsModel by just passing it a List containing a single ResultsModel
> * It doesn't take a claims list to build the Links collection - you'll need to just do this yourself in the facade if necessary... although I don't anticipate we will often need claims when building ReportReturnResources
> * I noticed a few places where a claims list was being passed needlessly to report facade services (and then seemingly not used?). Ripped them all out, but double check and let me know if you disagree

I know this should have been yesterday's conversation but why not just share an overload on ReportReturnResourceBuilder.Build that takes a single ResultsModel rather than having to force them all into lists?

lewisrenfrew:
> > * not quite finished but PR'ing now cos this could get ugly with merge conflicts...
> > * rewrote the ResultsModelResourceBuilder (and renamed it to ReportReturnResourceBuilder cos thats what it really is) in common. It now handles Lists of ResultsModels as opposed to a sole ResultsModel. You can make it handle a single ResultsModel by just passing it a List containing a single ResultsModel
> > * It doesn't take a claims list to build the Links collection - you'll need to just do this yourself in the facade if necessary... although I don't anticipate we will often need claims when building ReportReturnResources
> > * I noticed a few places where a claims list was being passed needlessly to report facade services (and then seemingly not used?). Ripped them all out, but double check and let me know if you disagree
> 
> I know this should have been yesterday's conversation but why not just share an overload on ReportReturnResourceBuilder.Build that takes a single ResultsModel rather than having to force them all into lists?

yeah, that;s a good idea

src/Facade/Services/PartsReceivedReportFacadeService.cs
```diff
@@ -27,14 +27,16 @@ public class PartsReceivedReportFacadeService : IPartsReceivedReportFacadeServic
...
             var resource = (ReportReturnResource)this.resultsModelResourceBuilder.Build(
-                this.domainService.GetReport(
-                    options.Jobref,
-                    options.Supplier,
-                    options.FromDate,
-                    options.ToDate,
-                    options.OrderBy,
-                    options.IncludeNegativeValues),
-                null);
+                new List<ResultsModel>
```
richardiphillips, line 30: list can go?

richardiphillips (approved):

PR #237, opened 2022-06-21
Update deliveries in one request

lewisrenfrew:
- previously an http PATCH was fired off for every row of the datagrid to be updated, but that sometimes meant firing 50 or so requests near simulatenously which made handling the responses a pain
- so I've refactored it to handle the updates in a single POST, which is very similiar to the csv spreadsheet upload post. The only difference really is the changes to be made are presented in a json array as opposed to  a csv file... which led me to:
- Made one batch update carter endpoint which correctly handles the update in both cases by looking at the content type of the request and handing off to the correct facade method based on whether it is text/csv or application/json.
- I had to implement a Nancy-esque  req.Bind< Resource> construct for this, since the implicit Carter resource binding (i.e. just passing in a Resource to the endpoint handler method) fell over since a json resource wasn't present in the text/csv case.
- That was a bit of a ramble - the code probably explains it better than this


src/Service/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    using System.IO;
+    using System.Threading.Tasks;
+
+    using Linn.Common.Serialization.Json;
+
+    using Microsoft.AspNetCore.Http;
+
+    public static class HttpRequestExtensions
+    {
+        public static async Task<T> Bind<T>(this HttpRequest req)
```
andrewfraser73, line 12: is this something we're going to need in common at some point?

andrewfraser73 (approved):
Think it looks okay

PR #240, opened 2022-06-29
More options

richardiphillips:
assemblies used on
where used select
more stock level options
2nd level part options
select by supplier as 1st and 2nd level options
stock locator link

### DIFF
--- a/src/Domain.LinnApps/MaterialRequirements/MrHeader.cs
+++ b/src/Domain.LinnApps/MaterialRequirements/MrHeader.cs
@@ -1,6 +1,7 @@
 ﻿namespace Linn.Purchasing.Domain.LinnApps.MaterialRequirements
 {
     using System.Collections.Generic;
+    using System.Security.Cryptography;
 
     public class MrHeader
     {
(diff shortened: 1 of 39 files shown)


src/Domain.LinnApps/MaterialRequirements/MrHeader.cs
```diff
@@ -84,6 +85,15 @@ public class MrHeader
...
 
+        public decimal? LatePurchaseOrders { get; set; }
+        
+        public string HighStockWithOrders { get; set; }
+
+        public string HighStockWithNoOrders { get; set; }
+
+        public string HasUnacknowledgedPurchaseOrders { get; set; }
+
+
```
lewisrenfrew, line 96: blanky

lewisrenfrew (approved):

src/Domain.LinnApps/MaterialRequirements/MrHeader.cs
richardiphillips: thanks. will fix in next branch

PR #243, opened 2022-06-29
Copyfiletooutput always + fix slashes

iaincrawf:


src/Facade/Services/PurchaseOrderFacadeService.cs
```diff
@@ -50,15 +50,15 @@ public string GetOrderAsHtml(int orderNumber)
         {
             var order = this.orderRepository.FindById(orderNumber);
 
-            var result = this.razorTemplateEngine.Render(order, "../Service.Host/views/" + @"\" + "PurchaseOrder.cshtml");
+            var result = this.razorTemplateEngine.Render(order, "..\\app\\views\\PurchaseOrder.cshtml");
```
richardiphillips, line 53: app view path should probably be an env variable?

richardiphillips (approved):

src/Facade/Services/PurchaseOrderFacadeService.cs
richardiphillips: ok i just read the teams thread and realised you're doing this anyway

PR #254, opened 2022-07-14
round to 4DP before checking prices match

lewisrenfrew:


iaincrawf (approved):
Cool, spotted a price mismatch test already there

PR #255, opened 2022-07-15
Purchase order editing (of a few fields)

iaincrawf:
They also want to edit some value fields, but I'm pretty sure we'll need to increase the totals fields based on that and also do currency conversion for the base value fields, so have left them disabled we've checked and added that functionality in

Writing domain tests for update just now


src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
```diff
@@ -398,20 +398,20 @@ public class PurchaseOrderDeliveryService : IPurchaseOrderDeliveryService
 
             detail.PurchaseDeliveries = newDeliveries.ToList();
 
-            // set mini order date requested to be first date requested of newly split deliveries
+            //// set mini order date requested to be first date requested of newly split deliveries
```
lewisrenfrew, line 401: weird double comment thingy

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -183,55 +183,162 @@ private void UpdateDetails(ICollection<PurchaseOrderDetail> currentDetails, ICol
...
+            miniOrder.InternalComments = updatedDetail.InternalComments;
+            miniOrder.SuppliersDesignation = updatedDetail.SuppliersDesignation;
+        }
+
+        /// <summary>
+        /// /below method currently unreferenced, but leaving
+        /// as it has most of what we'll need to put into mini order when creating
+        /// so will save re-doing it when we build create
+        /// </summary>
+        /// <param name="order"></param>
```
lewisrenfrew, line 261: I would just stick this in a notepad or something as opposed to merging unreferenced code

src/Resources/PurchaseOrderDeliveryResource.cs
```diff
@@ -57,5 +57,7 @@ public class PurchaseOrderDeliveryResource
         public string FilCancelled { get; set; }
 
         public decimal? QtyPassedForPayment { get; set; }
+
+        public EmployeeResource ConfirmedBy { get; set; }
```
lewisrenfrew, line 61: Was this used anywhere?

lewisrenfrew (approved):

src/Resources/PurchaseOrderDeliveryResource.cs
iaincrawf: Not so far no but it must be used somewhere.. no idea where

PR #257, opened 2022-07-18
refactor: abstract razor dependency from domain

lewisrenfrew:
- uses the ITemplateEngine abstraction I just tweaked in common


richardiphillips (approved):
think i'll need a show and tell to understand this one

iaincrawf (approved):
nice one, danke!

PR #259, opened 2022-07-19
Shortages : guard against no results, fix dropdown

RossStewart:
Misspoke in stand up, it is the vendor manager name (just for each planner). Just a small PR to fix a couple of gaps.


tests/Integration/Integration.Tests/ShortagesReportModuleTests/WhenGettingReportNoResults.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    using Linn.Common.Reporting.Models;
+    using Linn.Common.Reporting.Resources.ReportResultResources;
+    using Linn.Purchasing.Integration.Tests.Extensions;
+
+    using NSubstitute;
+
+    using NUnit.Framework;
+
+    public class WhenGettingReportNoResults : ContextBase
```
lewisrenfrew, line 17: I don't really think you need an extra carter/integration test for this scenario since its only your domain service you've changed, which is mocked out anyway

lewisrenfrew (approved):

tests/Integration/Integration.Tests/ShortagesReportModuleTests/WhenGettingReportNoResults.cs
RossStewart: No worries, will remove then. Thanks for the reminder!

lewisrenfrew: if you really felt the urge to test the no results scenario you could do a unit test for your domain service instead

PR #264, opened 2022-07-21
try logging email success and failures

iaincrawf:
not sure how to find the logs yet to check but will send some on sys with this there and see

- not sure if I'll get an actual error thrown if the email service fails, anyone know? 

- Does anyone know what I'm meant to put in the list of LoggingProperty? Fields I want logged like supplier and email maybe? I just put them in the message

### DIFF
--- a/src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
+++ b/src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
@@ -156,13 +156,13 @@
                     pdf.Result,
                     $"LinnPurchaseOrder{orderNumber}");
 
+            order.SentByMethod = "EMAIL";
+
+            //// todo When remove mini orders: stop setting sentbymethod mini order and switch to below
             var miniOrder = this.miniOrderRepository.FindById(orderNumber);
             miniOrder.SentByMethod = "EMAIL";
-            //// mini order trigger will update pl order. When remove mini orders:
-            //// todo stop setting sentbymethod mini order and switch to below
-            //// order.SentByMethod = "EMAIL";
 
-            return new ProcessResult(true, "Email Sent");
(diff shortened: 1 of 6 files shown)


tests/Unit/Facade.Tests/PurchaseOrderServiceTests/ContextBase.cs
```diff
@@ -45,6 +46,9 @@ public class ContextBase
 
         protected IFileReader FileReader { get; private set; }
 
+        protected ILog Logger { get; private set; }
+
+
```
RossStewart, line 51: blank

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -156,13 +156,13 @@ public ProcessResult SendPdfEmail(string html, string emailAddress, int orderNum
                     pdf.Result,
                     $"LinnPurchaseOrder{orderNumber}");
 
+            order.SentByMethod = "EMAIL";
+
+            //// todo When remove mini orders: stop setting sentbymethod mini order and switch to below
```
RossStewart, line 161: leaving the to do in?

RossStewart (approved):
looks good!

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
iaincrawf: yeah, can't remove mini order for a while! Unless you mean the actual todo, in which case yeah happy to leave it here just so it's super obvious, the rest are in their own methods but haven't done it here for updating 1 thing. Could but doesn't seem worth it. Although I changed the code so it doesn't actually make sense will update it haha

RossStewart (approved):

PR #265, opened 2022-07-22
fix PO dep and nom updating

iaincrawf:
I was previously including the nominal acc and dep/nom from that all in the po object, but EF was then trying to update the actual department/add a record and breaking.
Now just getting the nominal account from the order posting resource builder, and only setting the accountId on the posting in the domain service update method.

Just not sure if it should really be a proxy call to get nominal account, but doesn't feel right to be doing that in the resource builder.

If you guys are happy with this, feel free to run it locally or test it on sys and put through


RossStewart (approved):
Looks good (just fixed the context base)

PR #269, opened 2022-07-29
Ack orders upload tweaks

lewisrenfrew:
- accepts uploads of split deliveries (i.e. multiple delivs for one order) so long as the number and sequence of each delivery matches the records on our system
- doesn't actually do any automatic delivery splitting - that's still a manual job for the purchasing team
- previosuly the uploader just skipped any lines for deliveries that had been split, so this might offer some improvements


iaincrawf (approved):
sounds good

PR #273, opened 2022-08-03
po qty and value field updating

iaincrawf:
horribleness

Not sure if it's ok to just update OrderQty based on the conv factor that's (going to need to be) set upon creation?

Also need to look at exchange rates, I've started pulling it in here but actually could be more sensible to just set it when the currency changes, as the exchange rate is actually stored on here unlike other forms where I had to pull it in every time.. I can do that when doing creation though, as the current form doesn't allow updating of currency so doesn't matter yet.


iaincrawf:
I've updated the orderQty from ourQty and the conv factor, not sure if that's right and if I should also be updating the order price the same way from our price?

richardiphillips (approved):
honestly no idea if this is right or not but seems like a WIP so I don't want to hold it up

PR #275, opened 2022-08-04
stop qty or value being 0 on po reqs

iaincrawf:
default to 1 for both and just stays at 1 if user tries to enter less.

+ let users change value fields while draft or auth wait state


RossStewart (approved):
Looks good!

PR #277, opened 2022-08-10
Delivery performance

richardiphillips:


tests/Integration/Integration.Tests/LedgerPeriodModuleTests/ContextBase.cs
```diff
@@ -0,0 +1,46 @@
...
+        protected ITransactionManager TransactionManager { get; set; }
+
+        protected IFacadeResourceService<LedgerPeriod, int, LedgerPeriodResource, LedgerPeriodResource> LedgerPeriodFacadeService { get; private set; }
+
+        [SetUp]
+        public void EstablishContext()
+        {
+            this.TransactionManager = Substitute.For<ITransactionManager>();
+
+            this.LedgerPeriodFacadeService = Substitute.For<IFacadeResourceService<LedgerPeriod, int, LedgerPeriodResource, LedgerPeriodResource>>();
```
lewisrenfrew, line 34: I prefer using a real FacadeService in my integration tests...

 although I am wary that by introducing different testing styles I am pushing things to be less standardised

richardiphillips: yeah. you're right. there's no reason I shouldn't use a real one in simple examples like these. I'll think about it next time.

PR #278, opened 2022-08-10
Fore cast reports etc

lewisrenfrew:
- unfinished, but parking until I know how purchasing want the reports to look
- most of the wiring up bits in place


lewisrenfrew:
travis npm ci is failing since my PC isn't producing a lockfile that it can understand....

PR #280, opened 2022-08-12
Delivery Performance Details

richardiphillips:


src/Service.Host/package.json
```diff
@@ -78,7 +78,7 @@
     "@date-io/moment": "^2.11.0",
     "@emotion/react": "^11.6.0",
     "@emotion/styled": "^11.6.0",
-    "@linn-it/linn-form-components-library": "^13.12.0",
+    "@linn-it/linn-form-components-library": "^13.14.0",
```
lewisrenfrew, line 81: this will pull in my new export button (as I assume you noticed since the tests broke!_. Might be worth double checking it works before merging if you haven't already

lewisrenfrew (approved):

src/Service.Host/package.json
richardiphillips: thanks. checked one and seems ok.

PR #285, opened 2022-08-16
add split deliveries dialog

lewisrenfrew:


richardiphillips (approved):
as we discussed I'd change everything called split delivery to just delivery in this form

PR #286, opened 2022-08-17
orders - quick create front end, generate order from supplier function

iaincrawf:


richardiphillips (approved):
as before i don't really know if this is right but probably best to get it merged for now

iaincrawf:
> as before i don't really know if this is right but probably best to get it merged for now

these bits aren't saving anything yet and not linked to yet, so should be safe enough but agreed be good to merge it!

PR #288, opened 2022-08-19
Fore cast reports etc

lewisrenfrew:
still unfinished  (waiting on supplier feedback for report formats), but at a nice merge-able point


richardiphillips (approved):
lg

Issue #192, opened 2022-05-05
Add part description to part receipts report


andrewfraser73:
Request from Jacki

andrewfraser73:
sorted

PR #293, opened 2022-08-22
Partdescreceipts

andrewfraser73:
add part description to report as per issue #192


lewisrenfrew (approved):

tests/Unit/Domain.LinnApps.Tests/PrefSupReceiptsReportTests/WhenGettingReportJustGBP.cs
```diff
@@ -60,8 +60,8 @@ public void ShouldReturnJustGBPCurrencyFields()
         {
             var row = this.results.Rows.FirstOrDefault();
```
richardiphillips, line 61: no need to get the row here. Just use GetGridTextValue(0, 5) etc

PR #300, opened 2022-08-24
remove morelinq as nothing is really using it

richardiphillips:


lewisrenfrew:
.DistinctBy() is used in a couple of places? Is that not a MoreLinq thing?

lewisrenfrew (approved):

richardiphillips:
> .DistinctBy() is used in a couple of places? Is that not a MoreLinq thing?

i think that was what it was originally there for but it's in dot net core 6 now for free

PR #301, opened 2022-08-25
Nuget updates

richardiphillips:


lewisrenfrew (approved):

tests/Integration/Integration.Tests/Extensions/EnumerableExtensions.cs
```diff
@@ -10,4 +10,4 @@ public static T Second<T>(this IEnumerable<T> items)
             return items.ElementAt(1);
         }
     }
-}
\ No newline at end of file
+}
```
lewisrenfrew, line 13: I'd be tempted to just delete this extension... I don't really see how calling .Second() as any easier than calling .ElementAt(2) (I realise you probably didn't write it... just since you're doing the lords work and tidying things up)

richardiphillips: You're probably right. I was trying not to be too intrusive but it is a bit spurious

PR #306, opened 2022-08-29
Forecast change and close suppliers fixes

lewisrenfrew:


lewisrenfrew:
> helperText is cool. I either hadn't realised or had forgotten we pass that through.
> 
> Does the error mean that they can never uplift forecasts? Never had any reason to in the past. Just wondering.

No they still can - it just turns the box / helper text red if they do a positive one to make it more shouty

richardiphillips:
> > helperText is cool. I either hadn't realised or had forgotten we pass that through.
> > Does the error mean that they can never uplift forecasts? Never had any reason to in the past. Just wondering.
> 
> No they still can - it just turns the box / helper text red if they do a positive one to make it more shouty

sounds good

PR #311, opened 2022-08-31
add link to order and order by order number

richardiphillips:


src/Domain.LinnApps/Reports/PurchaseOrdersReportService.cs
```diff
@@ -390,61 +416,55 @@ private static void AddPartReportColumns(SimpleGridLayout reportLayout)
...
-                    });
 
             if (isFirstDelivery)
             {
+                values.Add(
+                    new CalculationValueModel
+                        {
+                            RowId = currentRowId,
+                            ColumnId = "Order",
+                            TextDisplay = $"{orderDetail.OrderNumber}"
```
iaincrawf, line 431: Will moving these inside this if() not only show values for the first delivery now? Previously was separate to show all deliveries, but only show qty fields for further deliveries on the same detail, so that the totals didn't include duplicate values like 874003 here for example 
http://app.linn.co.uk/purchasing/reports/orders-by-part/report?partNumber=RAW%20251&fromDate=2022-05-01T10:55:16.000Z&toDate=2022-08-31T10:55:16.536Z&cancelled=N

iaincrawf: Ignore this actually, think was just github's weird formatting, the whole file makes more sense

iaincrawf (approved):

PR #314, opened 2022-08-31
only check vs order unit price, not every deliveries unit price

lewisrenfrew:


richardiphillips:
do purchasing keep the detail price up to date then?

lewisrenfrew:
> do purchasing keep the detail price up to date then?

I think so. I hope so.

richardiphillips (approved):

PR #315, opened 2022-09-01
Monthly forecast report

lewisrenfrew:
- hid all the complexity of generating the report in a view, so pretty simple PR in the end just to craft the .csv from some repo calls
- still need to make an update to the common IEmailService before I merge to make it handle csv's again (currently it just takes a ResultsModel and does the .CoverToCsvList() internally


src/Domain.LinnApps/Mailers/SupplierAutoEmailsMailer.cs
```diff
@@ -107,9 +108,9 @@ private void CheckEmailDetailsOk(string toAddress, Supplier supplier)
             var lastTqmsDate = this.mrMaster.GetRecord().RunDate;
 
             // notify the vendor managers if tqms jobs failed
-            if (lastTqmsDate != DateTime.Today.Date)
+            if (lastTqmsDate.Date != DateTime.Today.Date)
```
richardiphillips, line 111: probably rename variable and change comment

richardiphillips (approved):
does this mean there isn't yet a displayable version of the report?

lewisrenfrew:
> does this mean there isn't yet a displayable version of the report?

Kind of. But there is a page where you can just download a csv which is what Andy said was preferable (theyd rather see exactly what the supplier sees which is just a csv)

PR #332, opened 2022-09-02
Monthly scheduling

lewisrenfrew:
- I'm not sure this works yet.. I tried to do some integration tests but not 100% convinced by them
- Will revisit on Monday but PR'ing now for that Friday PR feeling

### DIFF
--- a/src/Scheduling.Host/Jobs/SupplierAutoEmailsScheduler.cs
+++ b/src/Scheduling.Host/Jobs/SupplierAutoEmailsScheduler.cs
@@ -5,6 +5,7 @@
     using Linn.Common.Scheduling.Triggers;
     using Linn.Purchasing.Domain.LinnApps.Suppliers;
     using Linn.Purchasing.Resources.Messages;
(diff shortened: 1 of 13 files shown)


src/Scheduling.Host/Jobs/SupplierAutoEmailsScheduler.cs
```diff
@@ -14,63 +15,99 @@ public class SupplierAutoEmailsScheduler : BackgroundService
...
         }
 
         // Dispatches a message to instruct various reports to be emailed to suppliers on Monday mornings 6am
         // emails configured here https://app.linn.co.uk/purch/planning/plautoem.aspx
         protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
-            
-            // every day at 6am
-            var trigger = new DailyTrigger(6);
+            this.Log.Info("Supplier Auto Emails Scheduler Running...");
```
richardiphillips, line 40: small L instead of big L?

richardiphillips (approved):
looks good

PR #335, opened 2022-09-06
Amend search on PurchOrder and PurchOrderReq

RossStewart:
Remove Typeaheads and allow Purch to just go to desired order.

Played around a bit with the PurchOrderReq and thought some helper text may look better than a tooltip? Feel free to voice opinion though

### DIFF
--- a/src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrdersSearch.js
+++ b/src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrdersSearch.js
@@ -37,7 +37,7 @@ function PurchaseOrdersSearch() {
         dispatch(purchaseOrdersActions.fetchState());
         dispatch(purchaseOrdersActions.search(''));
     }, [dispatch]);
-    const searchResults = useSelector(state =>
+    const recentOrders = useSelector(state =>
         collectionSelectorHelpers.getSearchItems(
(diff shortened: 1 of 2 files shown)


src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrdersSearch.js
```diff
@@ -25,34 +46,71 @@ function PurchaseOrdersSearch() {
             'orderNumber'
         )
     );
-    const searchLoading = useSelector(state =>
-        collectionSelectorHelpers.getSearchLoading(state.purchaseOrders)
-    );
+    const [orderNumber, setOrderNumber] = useState('');
+    const search = () => history.push(`/purchasing/purchase-orders/${orderNumber}/`);
```
lewisrenfrew, line 50: if the search function just takes you straight to an order, then how do you ever get a searchResults list to render on this page (like you've started rendering on line 87

RossStewart: I'm using the collectionSelectorHelper's getSearchItems method on line 40 based on the order number and it's limited to 100 results. Maybe the naming could be a little better.

lewisrenfrew (approved):

PR #339, opened 2022-09-07
Create po backend

iaincrawf:
for now only works via /quick-create url first due to not all fields being editable on create yet, but could easily link to main create page from MR as long as we make a post to 'generate-order-fields' to fill out the fields from the supplier etc on the way to the create form

Scared I've missed bits, not sure whether to give purchasing a link to it yet or just put through for us to continue dev work


src/Facade/Services/PurchaseOrderFacadeService.cs
```diff
@@ -169,14 +175,27 @@ public string GetOrderAsHtml(int orderNumber)
             return this.domainService.GetPurchaseOrderAsHtml(orderNumber);
         }
 
+        public new IResult<PurchaseOrderResource> Add(PurchaseOrderResource resource, IEnumerable<string> privileges = null, int? userNumber = null)
+        {
+            var order = this.BuildEntityFromResourceHelper(resource);
+
+            this.domainService.CreateOrder(order, privileges);
```
lewisrenfrew, line 182: This CreateOrder( ) isn't really creating the order anymore, so maybe needs a different name

lewisrenfrew (approved):

src/Facade/Services/PurchaseOrderFacadeService.cs
iaincrawf: yeah I was actually kinda thinking that while I was looking at the facade earlier

PR #341, opened 2022-09-08
Accept different date format

lewisrenfrew:


andrewfraser73 (commented):
lgtm we're firmly in the 21st century now

PR #342, opened 2022-09-08
Print po

iaincrawf:
- print button on purchase order form which brings up print preview dialog
- In background there's a use effect when the order is loaded to get the html link and do a fetch, then put the html into a div that only shows during printing
- I did use dangerouslySetInnerHTML but I don't think there's much potential danger to us here and can't get it render the styling etc otherwise


src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
```diff
@@ -98,6 +99,19 @@ function PurchaseOrderUtility({ creating }) {
         }
     }, [item, applicationState, creating, reduxDispatch]);
 
+    const [printHtml, setPrintHtml] = useState(<span>loading</span>);
+
+    useEffect(() => {
+        if (item?.orderNumber) {
+            fetch(config.appRoot + utilities.getHref(item, 'html'))
```
iaincrawf, line 106: config approot bit is so that it works locally as it needs to hit the backend but otherwise gets swallowed by localhost3000

src/Service.Host/client/src/components/PurchaseOrders/POPrintout.js
```diff
@@ -0,0 +1,182 @@
+// import React, { useState, useEffect, useMemo } from 'react';
```
lewisrenfrew, line 1: Delete?

src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
```diff
@@ -346,190 +359,199 @@ function PurchaseOrderUtility({ creating }) {
     const getDateString = isoString =>
         isoString ? new Date(isoString).toLocaleDateString('en-GB') : null;
 
+    const print = () => {
+        window.print();
```
lewisrenfrew, line 363: there's probably a more react-y way to do this.. have you ever used react to print? https://www.npmjs.com/package/react-to-print

lewisrenfrew (approved):

src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
lewisrenfrew: hooks-y example here https://github.com/linn/stores/blob/master/src/Service.Host/client/src/components/tpk/Tpk.js

src/Service.Host/client/src/components/PurchaseOrders/POPrintout.js
iaincrawf: oops thought I already had

src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
iaincrawf: I can if ya'd rather, but almost seems sort of backwards to change a line of code to several lines and add complexity with react to print. I guess accessing window isn't too react-y, but it doesn't feel entirely wrong to me (though maybe it could be in terms of react purity). I was considering getting rid of the function and just having () => window.print() in the onClick

iaincrawf: yeah that side of it does look quite nice, and hadn't really looked into it before either though so good to be aware of it :)

PR #359, opened 2022-09-15
Purchase order link from MR + fixes/improvements

iaincrawf:
- link from MR
- Bug fixes
- set more fields by default:
- unit of measure, nominal/dept from part supplier after wizard step in domain fill out method, IssuePartsToSupplier based on part.SupplierAssembly()
- lead time to push advised date (or should it be requested date) out


andrewfraser73 (approved):
SOTM

PR #365, opened 2022-09-20
po req check

richardiphillips:


src/Facade/ResourceBuilders/CurrencyResourceBuilder.cs
```diff
@@ -4,17 +4,28 @@
...
+        private readonly ICurrencyPack currencyPack;
+
+        public CurrencyResourceBuilder(ICurrencyPack currencyPack)
+        {
+            this.currencyPack = currencyPack;
+        }
+
         public CurrencyResource Build(Currency entity, IEnumerable<string> claims)
         {
+            var exchangeRate = this.currencyPack.GetExchangeRate(entity.Code, "GBP");
```
iaincrawf, line 22: ooh hadn't thought of this in the resource builder, saves a lot of messing about

iaincrawf (approved):
nice one, thanks for fixing

PR #367, opened 2022-09-21
Bom tree ++

lewisrenfrew:
traverse the bom tree client-side so we can do each branch in parallel and load big trees quickly


andrewfraser73 (approved):
seems okay - could lead to lots of connections but we'll see

PR #368, opened 2022-09-21
Add MR comments to Shortages Planner Report per Item

RossStewart:
Updated the View to include the MR comments. Minor changes to Planner report to include the comments on the end of the item.


src/Domain.LinnApps/Reports/ShortagesReportService.cs
```diff
@@ -127,6 +127,7 @@ public IEnumerable<ResultsModel> GetShortagesPlannerReport(int planner)
                     model.AddColumn("TotalBTReqt", "To achieve Ideal Build");
                     model.AddColumn("Supplier", "Supplier");
                     model.AddColumn("SupplierName", "Supplier Name");
+                    model.AddColumn("PlannerStory", "Planner Story");
```
richardiphillips, line 130: Don't call it planner story on the actual report. That's just what the table column happens to be called but they won't know what it is

richardiphillips (approved):
do you need me to add planner_story to the database view?

src/Domain.LinnApps/Reports/ShortagesReportService.cs
RossStewart: Noted! Will change it

RossStewart:
> do you need me to add planner_story to the database view?

all sorted!

PR #384, opened 2022-09-30
PO Changes. Searches and stock controlled parts.

RossStewart:


src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -198,9 +202,24 @@ function CreatePurchaseOrderUt() {
...
+        stockControlled: c.stockControlled
     }));
 
+    const partSuppliersSearchResults = useSelector(reduxState =>
+        collectionSelectorHelpers
+            .getSearchItems(reduxState.partSuppliers)
+            ?.sort((a, b) => (a.supplierRanking ?? 100) - (b.supplierRanking ?? 100))
+            .map(s => ({
+                id: s.supplierId,
+                displayText: `${s.supplierName} RANK : ${s.supplierRanking}`
```
lewisrenfrew, line 215: I don't think you need to display supplierRanking here (its often null. Just indicating which one has supplier ranking = 1 (i.e. which is the preferred supplier) is probably better.

lewisrenfrew (changes requested):

richardiphillips:
Looks like you're updating a very out of date file. Definitely merge main back in before testing anything. 

You really shouldn't be sitting in branches for this length of time as it just causes yourself problems.

RossStewart:
Yeah, it's bit a of merge. Merged back main and just checking locally first before anything.

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -161,9 +164,20 @@ function CreatePurchaseOrderUt() {
...
+        stockControlled: c.stockControlled
     }));
 
+    const partSuppliersSearchResults = useSelector(reduxState =>
+        collectionSelectorHelpers.getSearchItems(reduxState.partSuppliers)
+    );
+
+    useEffect(() => {
+        if (partSuppliersSearchResults?.length) {
+            handleSupplierChange(partSuppliersSearchResults[0]);
```
lewisrenfrew, line 177: I'd fish out the preferred supplier here as opposed to just choosing the first one

lewisrenfrew (approved):

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
RossStewart: Mm this was the original purpose of the sort that was implemented originally on the declaration of partSuppliersSearchResults (now just in the FetchItems). How else would I tell the preferred supplier other than choosing the first item of the sorted list (by supplier ranking)?

lewisrenfrew: partSupplierSearchResults won't be sorted by supplierRanking now since we moved the sort into the Dropdowns items list.

I'd just use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

PR #404, opened 2022-10-10
Chgbomtype server side

andrewfraser73:
this is just the server side part of chgbomtype 

it provides a POST endpoint to do a bom_Type change on Parts and make a part history record

I refactored some of the common logic between preferred supplier change and bom type change into separate services


tests/Unit/Domain.LinnApps.Tests/PartSupplierServiceTests/WhenCreatingAPreferredSupplierChangeForAPartForTheFirstTime.cs
```diff
@@ -127,27 +125,7 @@ public void ShouldUpdatePartSupplierRecords()
...
-                && x.ChangeType == "PREFSUP"
-                && x.Remarks == "REMARKS"
-                && x.PriceChangeReason == "CHG"
-                && x.OldCurrency == null
-                && x.NewCurrency == "USD"
-                && x.OldCurrencyUnitPrice == null
-                && x.NewCurrencyUnitPrice == 100m
-                && x.OldBaseUnitPrice == null
-                && x.NewBaseUnitPrice == 3m));
+            this.PartHistory.Received().AddPartHistory(Arg.Any<Part>(), Arg.Any<Part>(), "PREFSUP", Arg.Any<int>(),"REMARKS","CHG");
```
lewisrenfrew, line 128: seems like this test is losing some specificity compared to the one its replacing. Not sure whether that really matters or not

lewisrenfrew (approved):

tests/Unit/Domain.LinnApps.Tests/PartSupplierServiceTests/WhenCreatingAPreferredSupplierChangeForAPartForTheFirstTime.cs
andrewfraser73: If I was worried about AddPartHistory I could probably write unit tests for that

PR #408, opened 2022-10-12
Misc tickets and bugs

lewisrenfrew:


src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -416,11 +416,31 @@ public PurchaseOrder FillOutUnsavedOrder(PurchaseOrder order, int currentUserId)
...
 
-            //TODO THIS IS WRONG AND ALSO HAS A DATABASE ID HARDCODED INTO DOMAIN LOGIC
-            // from MR is always nom Raw Materials 0000007617 Assets 0000002508
-            var nomAcc = this.nominalAccountRepository.FindById(884);
+            NominalAccount nomAcc = null;
+            if (part.StockControlled == "Y")
+            {
+                if (part.RawOrFinished == "R")
+                {
+                    nomAcc = this.nominalAccountRepository.FindById(884);
```
richardiphillips, line 424: i think these should find by nominal and department codes rather than database id. The domain shouldn't know anything about the data storage (and would fail if we built our own development databases).

richardiphillips (approved):

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
lewisrenfrew: good call

PR #409, opened 2022-10-12
tiny layout hotfix to stop supplier designation falling down

richardiphillips:


lewisrenfrew (approved):
oops - i must have broke this earlier. sorry!

PR #411, opened 2022-10-13
set type and orig order number

richardiphillips:


src/Domain.LinnApps/PurchaseOrders/PlCreditDebitNoteService.cs
```diff
@@ -191,11 +191,10 @@ public void CreateDebitOrNoteFromPurchaseOrder(PurchaseOrder order)
                     CancelledBy = null,
                     DateCancelled = null,
                     ReasonCancelled = null,
-                    NoteType = order.DocumentTypeName == "CO"
```
lewisrenfrew, line 194: I have no idea why I did this and worrryingy can't remember doing it at all - Apologies if it cause Andy & co problems

Issue #401, opened 2022-10-06
PO : Requested Date


RossStewart:
Andy wants to see the requested date on the Create screen. 

This should default to the lead time + delivery day e.g. if a Supplier has a delivery date of thursday, it would be leadtime + 4 days. 

If a supplier has no delivery date, it should default to 2 weeks.

PR #414, opened 2022-10-14
Requested date business

lewisrenfrew:
for #401 
- all works fine except date pickers show the wrong day of the week


RossStewart (approved):
Looks good!

PR #419, opened 2022-10-18
Our price currency op

lewisrenfrew:


src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -390,7 +399,7 @@ function CreatePurchaseOrderUt() {
                                     handleDetailFieldChange(
                                         'partNumber',
                                         newPart.id,
-                                        order.details[0]
+                                        order.details[0].id
```
lewisrenfrew, line 402: I think this is the only problem now: handleDetailFieldChange still takes the whole detail as its third arg, i.e. not the id

PR #420, opened 2022-10-18
Our Currency + Qty Price Fix

RossStewart:
This led us down several wrong paths but after testing this, I believe this is it.

### DIFF
--- a/src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
+++ b/src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
@@ -111,8 +111,11 @@ function CreatePurchaseOrderUt() {
 
     const canSave = () => allowedToCreate() && inputIsValid();
 
-    const handleDetailFieldChange = (propertyName, newValue, detail) => {
-        dispatch({ payload: { ...detail, [propertyName]: newValue }, type: 'detailFieldChange' });
+    const handleDetailFieldChange = (propertyName, newValue, detailId) => {
+        dispatch({
+            payload: { id: detailId, [propertyName]: newValue },
+            type: 'detailFieldChange'
+        });
     };
 
     const handleDetailValueFieldChange = (propertyName, basePropertyName, newValue, detail) => {
@@ -171,14 +174,6 @@ function CreatePurchaseOrderUt() {
     const partSuppliersSearchResults = useSelector(reduxState =>
         collectionSelectorHelpers.getSearchItems(reduxState.partSuppliers)
     );
-
-    useEffect(() => {
-        if (partSuppliersSearchResults?.length) {
-            handleSupplierChange({
-                id: `${partSuppliersSearchResults.find(s => s.supplierRanking === 1).supplierId}`
-            });
-        }
-    }, [partSuppliersSearchResults]);
 
     const partsSearchLoading = useSelector(state =>
         collectionSelectorHelpers.getSearchLoading(state.parts)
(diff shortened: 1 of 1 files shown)


src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -278,6 +270,23 @@ function CreatePurchaseOrderUt() {
...
+                payload: {
+                    lineNumber: 1,
+                    fieldName: 'ourUnitPriceCurrency',
+                    value: partSuppliersSearchResults.find(s => s.supplierRanking === 1)
+                        .currencyUnitPrice
+                },
+                type: 'detailFieldUpdate'
+            });
+        }
+    }, [detail.id, partSuppliersSearchResults]);
```
lewisrenfrew, line 288: can you remove detail.id from this array? It doesn't seem to be used in the effect

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -454,6 +463,14 @@ function CreatePurchaseOrderUt() {
                                         }))}
                                     onChange={(propertyName, selected) => {
                                         handleSupplierChange({ id: selected });
+                                        handleDetailValueFieldChange(
```
lewisrenfrew, line 466: I would change this to be the same dispatch as the one above for consistency. (i.e. dispatch an action of type detailFieldUpdate in both cases)

lewisrenfrew (approved):

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
RossStewart: This is one of the functions that has some currency conversion before the actual dispatch. I think for uniformity sake, it may stick out if I change it? There's another instance of HandleDetailValueFieldChange within the code as well.

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
RossStewart: sorted.

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -278,6 +252,23 @@ function CreatePurchaseOrderUt() {
...
+                payload: {
+                    lineNumber: 1,
+                    fieldName: 'ourUnitPriceCurrency',
+                    value: partSuppliersSearchResults.find(s => s.supplierRanking === 1)
+                        .currencyUnitPrice
+                },
+                type: 'detailFieldUpdate'
+            });
+        }
+    }, [detail.id, partSuppliersSearchResults]);
```
lewisrenfrew, line 270: detail.id is still on here

src/Service.Host/client/src/components/PurchaseOrders/CreatePurchaseOrderUt.js
```diff
@@ -526,12 +527,14 @@ function CreatePurchaseOrderUt() {
...
-                                    handleDetailValueFieldChange(
-                                        propertyName,
-                                        'baseUnitPrice',
-                                        newValue,
-                                        detail
-                                    )
+                                    dispatch({
+                                        payload: {
+                                            lineNumber: 1,
+                                            fieldName: 'propertyName',
```
lewisrenfrew, line 533: this will break this InputField since there is no property with fieldName 'propertyName' on a purchaseOrderDetail.

PR #436, opened 2022-10-31
Change SentByMethod from InputField to Dropdown

RossStewart:


richardiphillips (approved):
should be a test too?

lewisrenfrew:
Does the domain service actually upate this field? I couldn't see where it did

lewisrenfrew:
> > Does the domain service actually upate this field? I couldn't see where it did

> Yeah, I'm seeing it in the Database. I just created a PO and if you look up Order 827542 you should see the method set to POST from this dropdown.

Where does it happen?

RossStewart:
> > > Does the domain service actually upate this field? I couldn't see where it did
> 
> > Yeah, I'm seeing it in the Database. I just created a PO and if you look up Order 827542 you should see the method set to POST from this dropdown.
> 
> Where does it happen?

SENT_BY_METHOD is referenced in the Build. It's not on an update, it's on the creation of the PO. 


Showed Andrew this earlier but I'll just double check - should we be allowing an edit after the PO has been created?

RossStewart:
> > > > Does the domain service actually upate this field? I couldn't see where it did
> > 
> > 
> > > Yeah, I'm seeing it in the Database. I just created a PO and if you look up Order 827542 you should see the method set to POST from this dropdown.
> > 
> > 
> > Where does it happen?
> 
> SENT_BY_METHOD is referenced in the Build. It's not on an update, it's on the creation of the PO.
> 
> Showed Andrew this earlier but I'll just double check - should we be allowing an edit after the PO has been created?
> 
> I'm not sure. That's not what I mean though. I'm just curious what line of code actually sets this field in the domain service when we create a PO. I can't see it. Can you write a test to check that your code is actually setting this field when creating a PO?

Yeah no problem. From what I can see we are doing a BuildEntityFromResourceHelper in the facade service (where the value has been set in the resource) and then that order is passed to the domain service to be created.

Asked Andrew if we should be able to update it after the fact and will make any changes that need to be done in here and re-request review as well as adding in any tests.

RossStewart:
They don't set the field when Creating, only when updating. Updated the update logic in. Added test in that checks both mini order and PO.

lewisrenfrew:
> Can set when creating or updating. Added the update logic in. Added test in that checks SentByMethod for both mini order and PO as well as the other fields that can be updated post creation.

cool - you should do a test for the creating case as well then. Actually you probably don't need a whole new test. Just add to this one to test the SentByMethod is set https://github.com/linn/purchasing/blob/main/tests/Unit/Domain.LinnApps.Tests/PurchaseOrderServiceTests/WhenCreatingPurchaseOrder.cs

PR #441, opened 2022-11-04
entry level board

richardiphillips:


andrewfraser73 (approved):
think you'll need PCAS_REVISIONS and PCAS_LAYOUTS before too long as just boards and components don't make sense w/o it

richardiphillips:
> think you'll need PCAS_REVISIONS and PCAS_LAYOUTS before too long as just boards and components don't make sense w/o it

yep. starting with baby steps

PR #445, opened 2022-11-08
Cl tools

lewisrenfrew:
- simple tool to let you send some or all of the scheduled order books emails (for real or to a test address) manually


richardiphillips (approved):
cool. should it be in its own folder in case we write other tools?

Maybe a wee readme.txt with how a correct command line would look for test and real?

PR #450, opened 2022-11-11
bcc to the acknowledgements box

lewisrenfrew:


richardiphillips (approved):
should probably add it to config example too

lewisrenfrew:
> should probably add it to config example too

yeah... I'm also debating moving all these email addresses to a database table

PR #453, opened 2022-11-11
Chgbomtype2

andrewfraser73:
UI and a bit more backend for bom type change

we have lots of validation checks on the server side trying to replicate most of that also on client


src/Facade/Services/PartFacadeService.cs
```diff
@@ -64,6 +64,21 @@ public string GetPartNumberFromId(int id)
             }
         }
 
+        public IResult<BomTypeChangeResource> GetBomType(int partNumberId, IEnumerable<string> privileges = null)
```
richardiphillips, line 67: can't you just get this information from the part without having a separate facade method for it?

src/Service.Host/client/src/middleware/itemCreated.js
```diff
@@ -5,7 +5,7 @@ import * as actionTypes from '../actions/index';
 export default () => next => action => {
     const result = next(action);
 
-    if (action.type.startsWith('RECEIVE_NEW_')) {
+    if (action.type?.startsWith('RECEIVE_NEW_')) {
```
richardiphillips, line 8: i'd revert this. actions should always have a type

src/Service.Host/client/src/components/BomTypeChange.js
```diff
@@ -0,0 +1,283 @@
...
+import ModeEditIcon from '@mui/icons-material/ModeEdit';
+import EditOffIcon from '@mui/icons-material/EditOff';
+import Tooltip from '@mui/material/Tooltip';
+import history from '../history';
+import config from '../config';
+import partsActions from '../actions/partsActions';
+import bomTypeChangeActions from '../actions/bomTypeChangeActions';
+import partSuppliersActions from '../actions/partSuppliersActions';
+
+function BomTypeChange() {
```
richardiphillips, line 28: is there a test for this?

richardiphillips (approved):

src/Facade/Services/PartFacadeService.cs
andrewfraser73: I think it could be a possible refactor unless i remember what pushed me down this route

PR #460, opened 2022-11-21
Add new Delivery when Qty field overridden

RossStewart:


tests/Unit/Domain.LinnApps.Tests/PurchaseOrderServiceTests/WhenUpdatingPurchaseOrderConvValueQtyFields.cs
```diff
@@ -486,7 +486,7 @@ public void ShouldUpdateSecondDeliveryTotalFields()
             var firstDetail = this.current.Details.First();
 
             // get second delivery
-            var delivery = firstDetail.PurchaseDeliveries.Last();
+            var delivery = firstDetail.PurchaseDeliveries.ElementAt(1);
```
richardiphillips, line 489: I think you should revert this test and add another for the scenario in question. Ideally you should add the new test before making any code changes

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -954,6 +954,42 @@ private void SetDetailFieldsForCreation(PurchaseOrderDetail detail, int newOrder
 
         private void UpdateDeliveries(PurchaseOrderDetail purchaseOrderDetail, decimal exchangeRate, int supplierId)
         {
+            if (purchaseOrderDetail.OrderQty > purchaseOrderDetail.OurQty)
```
richardiphillips, line 957: I don't understand this? I thought from what Andrew said this is to do with keeping deliveries in sync with the order line. I don't think it's anything to do with order qty and our qty. 

Also it presumably can't just be about adding a delivery. What about the scenario where they reduce the qty (our qty) on the line?

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderServiceTests/WhenUpdatingPurchaseOrderConvValueQtyFields.cs
richardiphillips: or perhaps two new tests in this scenario?

richardiphillips (changes requested):

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
richardiphillips: I haven't seen the ticket so this is maybe a different issue from the one that Andrew mentioned this morning?

richardiphillips: It doesn't seem to address the 0 value problem either

RossStewart: Andrew said that they never reduce the qty, it's just not a thing that they do. Asked about the scenario but he was pretty certain it shouldn't ever happen.

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderServiceTests/WhenUpdatingPurchaseOrderConvValueQtyFields.cs
RossStewart: That's fair, I just figured that it was an extension of the current functionality so could just add to the test but it's fine to add a completely new one.

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
richardiphillips: maybe they shouldn't but it looks like that's what's happened in the example that Andrew posted. 
If it's not supposed to happen then we should disallow it

RossStewart: Yeah I can change this on the front end. @lewisrenfrew Any thoughts on this? I know we discussed a few different approaches.

lewisrenfrew: I would just throw an exception

RossStewart: Yeah, ok definitely. General approach lines up with what you expected though with the use of the Qty fields?

RossStewart:
Changes submitted. React test for form coming on different PR.

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsAddingToDeliveryTotal.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        [Test]
+        public void ShouldReturnResult()
+        {
+            //Delivery fields
+            this.result.Count().Should().Be(4);
+            var updateData = this.updatedDeliveries.First();
+            var updated = this.result.First();
+            updated.DeliverySeq.Should().Be(1);
```
lewisrenfrew, line 123: Most of this stuff should already be covered in existing tests so you don't need to test it all again. I would just test that the details quantity is the correct value (along with the monetary totals that depend on the quantity

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsRemovingFromDeliveryTotal.cs
```diff
@@ -0,0 +1,138 @@
...
+
+
+            // Order will be updated to delivery total. Simulate removing deliveries.  
+            this.line = new PurchaseOrderDetail
+            {
+                Line = 1,
+                OrderUnitPriceCurrency = 666,
+                OrderConversionFactor = 1.5m,
+                BaseOurUnitPrice = 111,
+                PurchaseDeliveries = new List<PurchaseOrderDelivery>(),
```
lewisrenfrew, line 57: If this test is for the scenario where you are removing deliveries, then shouldn't this list have more deliveries in it than the 'updatedDeliveries' list?

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsRemovingFromDeliveryTotal.cs
```diff
@@ -0,0 +1,138 @@
...
+            //Delivery fields
+            this.result.Count().Should().Be(1);
+            var updateData = this.updatedDeliveries.First();
+            var updated = this.result.First();
+            updated.DeliverySeq.Should().Be(1);
+            updated.OurDeliveryQty.Should().Be(updateData.OurDeliveryQty);
+            updated.OrderDeliveryQty.Should().Be(updateData.OurDeliveryQty / this.line.OrderConversionFactor);
+            updated.OurUnitPriceCurrency.Should().Be(this.line.OurUnitPriceCurrency);
+            updated.DateRequested.Should().Be(updateData.DateRequested);
+            updated.DateAdvised.Should().Be(updateData.DateAdvised);
```
lewisrenfrew, line 103: again you're needlessly testing a whole bunch of fields here that are irrelevant to the scenario of adding/removing deliveries. I'd just test the ones that matter (order line qty and related total values)

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsRemovingFromDeliveryTotal.cs
```diff
@@ -0,0 +1,138 @@
...
+            this.MiniOrderRepository.FindById(this.order.OrderNumber).Returns(new MiniOrder { });
+            this.result = this.Sut.UpdateDeliveriesForOrderLine(
+                this.order.OrderNumber,
+                this.line.Line,
+                this.updatedDeliveries,
+                new List<string>());
+        }
+
+        [Test]
+        public void ShouldReturnResult()
```
lewisrenfrew, line 92: Test name could be better again

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsAddingToDeliveryTotal.cs
```diff
@@ -0,0 +1,163 @@
...
+                                                     DeliverySeq = 1,
+                                                     OurDeliveryQty = 10,
+                                                     BaseOurUnitPrice = 11,
+                                                     DateRequested = 25.March(2022),
+                                                     AvailableAtSupplier = "N"
+                                                 },
+
+                                         };
+
+
```
lewisrenfrew, line 74: blank line

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsRemovingFromDeliveryTotal.cs
```diff
@@ -0,0 +1,138 @@
...
+                                                 {
+                                                     DeliverySeq = 1,
+                                                     OurDeliveryQty = 10,
+                                                     BaseOurUnitPrice = 11,
+                                                     DateRequested = 25.March(2022),
+                                                     AvailableAtSupplier = "N"
+                                                 }
+                                         };
+
+
```
lewisrenfrew, line 49: blank

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingDetailsRemovingFromDeliveryTotal.cs
```diff
@@ -0,0 +1,138 @@
...
+            updated.AvailableAtSupplier.Should().Be(updateData.AvailableAtSupplier);
+
+            //Sum total of the delivery quantities
+            line.OrderQty.Should().Be(10);
+        }
+
+        [Test]
+        public void ShouldUpdateMiniOrder()
+        {
+            this.MiniOrderRepository.Received().FindById(this.order.OrderNumber);
```
lewisrenfrew, line 135: one last thing - the UpdateMiniOrder() method in the PurchaseOrderDeliveryService doesn't actually update the qty on the mini order, so as it stands this will fall out of sync with the  PL_ORDER unless you change it. Are we accepting that mini_order is just to die now? If not you should make sure you update the mini order qty in step with the pl_order_detail qties too

richardiphillips: i think we need to keep it in sync for a bit longer

RossStewart: I think that's a question for Richard but I'll keep it in sync for now

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingShouldDecreaseOrderLineQty.cs
```diff
@@ -0,0 +1,132 @@
...
+                                         };
+
+            // Order will be updated to delivery total.
+            this.line = new PurchaseOrderDetail
+            {
+                Line = 1,
+                OrderUnitPriceCurrency = 666,
+                OrderConversionFactor = 1.5m,
+                BaseOurUnitPrice = 111,
+                PurchaseDeliveries = new List<PurchaseOrderDelivery>(),
```
lewisrenfrew, line 56: this list needs to have more deliveries in it than the current lines deliveries for this test to make sense

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
```diff
@@ -535,6 +535,8 @@ public class PurchaseOrderDeliveryService : IPurchaseOrderDeliveryService
 
             detail.PurchaseDeliveries = newDeliveries.ToList();
 
+            this.purchaseOrderService.UpdateOrder(order, order, privileges);
+
             // set mini order date requested to be first date requested of newly split deliveries
             this.UpdateMiniOrder(
```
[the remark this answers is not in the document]
RossStewart: Qty on the mini order is getting updated through this updateorder route. Can confirm it by running it then running the following 
'select * from MINI_ORDER where ORDER_NUMBER = XXXXXX'

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
```diff
@@ -535,6 +535,8 @@ public class PurchaseOrderDeliveryService : IPurchaseOrderDeliveryService
 
             detail.PurchaseDeliveries = newDeliveries.ToList();
 
+            this.purchaseOrderService.UpdateOrder(order, order, privileges);
```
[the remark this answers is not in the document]
RossStewart: It's the only really Public way through to that part of the service. It definitely wasn't designed for this. Is there any other way I should be getting to it? I could make some other functions in that service Public but didn't think that was a great idea. I do get that this isn't the most readable thing i nthe world though.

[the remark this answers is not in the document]
RossStewart: This is due to making use of the PurchaseOrderService and making sure that the detail calculations are being performed there rather than writing the same function as well a few other parts of that service in the deliveries. That could be an alternative or I could open up a more direct route?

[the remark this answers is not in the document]
lewisrenfrew: ooh I see what you mean. If it works it works - A lot of this will probably end up getting refactored anyways

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderDeliveryService.cs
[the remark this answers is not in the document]
lewisrenfrew: interesting... maybe a database trigger is taking care of it? thats fine anyways

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderDeliveryServiceTests/WhenUpdatingShouldDecreaseOrderLineQty.cs
```diff
@@ -0,0 +1,132 @@
...
+            updated.OurDeliveryQty.Should().Be(updateData.OurDeliveryQty);
+            updated.OrderDeliveryQty.Should().Be(updateData.OurDeliveryQty / this.line.OrderConversionFactor);
+            updated.NetTotalCurrency.Should().Be(
+                Math.Round(
+                    updateData.OurDeliveryQty.GetValueOrDefault() * this.line.OurUnitPriceCurrency.GetValueOrDefault(),
+                    2));
+            updated.VatTotalCurrency.Should().Be(this.vatAmount);
+            updated.DeliveryTotalCurrency.Should().Be(
+                Math.Round(
+                    this.line.OrderUnitPriceCurrency.GetValueOrDefault() * updated.OurDeliveryQty.GetValueOrDefault(),
```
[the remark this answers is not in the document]
lewisrenfrew: I think this is ok actually on second thoughts after reading your other comments. Since it so happens that there is only one delivery the sum all quantities is just that first quantity and everything checks out ok

lewisrenfrew (approved):

src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
```diff
@@ -1198,83 +1197,29 @@ function PurchaseOrderUtility({ creating }) {
...
                                                             )
                                                         }
-                                                        disabled={!allowedToUpdate()}
+                                                        disabled={!creating}
                                                         type="number"
                                                         required
                                                     />
                                                 </Grid>
                                                 <Grid item xs={6}>
-                                                    {overridingOrderQty ? (
```
richardiphillips, line 1207: Is there a reason that this has been removed? Surely it's still needed in create mode if the qties are still set here at that time?

richardiphillips (approved):

PR #465, opened 2022-11-25
Layouts

richardiphillips:


src/Service.Host/client/src/components/boards/boardReducer.js
```diff
@@ -0,0 +1,133 @@
+const initialState = {
+    selectedLayout: null,
+    board: { coreBoard: 'N', clusterBoard: 'N', idBoard: 'N', splitBom: 'N', layouts: [] }
+};
+
+const layoutCodesAreUnique = (layouts, newLayoutCode) => {
+    if (!layouts && !layouts.length) {
```
lewisrenfrew, line 7: isn't the second part of this && statement always true if the first part is?

lewisrenfrew (approved):

src/Service.Host/client/src/components/boards/boardReducer.js
richardiphillips: probably. will make it more succinct next

PR #467, opened 2022-11-29
Bom costs

lewisrenfrew:
Work in progress... general approach is: 
- use the BomTreeService.FlattenBomTree() method to get all the DetailId's on the bom
- query a new database view to get all the associated costing information for that detail efficiently

I haven't actually calculated the totals yet since I need some user input first. Will do that and add the corresponding tests in the next PR to make sure the boms are being costed right

Also need to ask about additional parameters and flesh them out or remove them as required

### DIFF
--- a/src/Persistence.LinnApps/Repositories/BomRepository.cs
+++ b/src/Persistence.LinnApps/Repositories/BomRepository.cs
@@ -45,7 +45,7 @@
         {
             return this.serviceDbContext.Boms.Where(expression)
                 .Include(b => b.Details)
-                .ThenInclude(d => d.Part)
+                .ThenInclude(d => d.Part).ThenInclude(p => p.PartSuppliers.Where(ps => ps.SupplierRanking == 1))
                 .Include(b => b.Details)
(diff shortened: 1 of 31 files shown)


src/Persistence.LinnApps/Repositories/BomRepository.cs
```diff
@@ -45,7 +45,7 @@ public Bom FindBy(Expression<Func<Bom, bool>> expression)
         {
             return this.serviceDbContext.Boms.Where(expression)
                 .Include(b => b.Details)
-                .ThenInclude(d => d.Part)
+                .ThenInclude(d => d.Part).ThenInclude(p => p.PartSuppliers.Where(ps => ps.SupplierRanking == 1))
```
andrewfraser73, line 48: is supplier ranking ==1more reliable that parts - > part_supplier via preferred_supplier?

andrewfraser73 (approved):
seems okay assuming you've compared some bom costs with the old report to check comes up with same figures

PR #469, opened 2022-11-30
don't allow price change after delivery

richardiphillips:


lewisrenfrew (approved):

lewisrenfrew:
could put in a c# check too i suppose but i appreciate the 'hot'ness of this hot fix

richardiphillips:
> could put in a c# check too i suppose but i appreciate the 'hot'ness of this hot fix

yeah. probably should have done that too

PR #473, opened 2022-12-01
Finish bom costs and other misc todos

lewisrenfrew:
- assorted tweaks resulting from meeting with jacki yesterdat

### DIFF
--- a/src/Domain.LinnApps/Boms/BomDetail.cs
+++ b/src/Domain.LinnApps/Boms/BomDetail.cs
@@ -39,5 +39,7 @@
         public Part BomPart { get; set; }
 
         public string PcasLine { get; set; }
+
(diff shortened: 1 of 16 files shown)


andrewfraser73 (approved):
approved although bom detail component is a weird marriage of convenience between bom_details and pcas_components but I guess there's no better name for it

lewisrenfrew:
> approved although bom detail component is a weird marriage of convenience between bom_details and pcas_components but I guess there's no better name for it

yeah ... It's a c# representation of a new view I created which is basically just the same as PCAS_REVISION_COMP_VIEW with the detail_id joined on. I had to have the detail_id to get my EFCore relationship to work. If you think there's a better way let me know

PR #482, opened 2022-12-08
Support tickets

lewisrenfrew:
- default to most recently selected date advised when acknowledging orders (as opposed to todays date)
- don't allow raising orders for closed suppliers
- replace typeaheads for new search components in purchase order quick-create screen - Let me know if you think thats a bad idea


richardiphillips (approved):
and I think Search is a very good idea. Had made a note to do this myself as the typeaheads annoyed me when I had to test it the other day

PR #485, opened 2022-12-09
Default fields

richardiphillips:


andrewfraser73 (approved):
i take it this to default the PCAS/PCSM/PCB to sensible defaults when creating.  at some point we may offer part creation at this step too.

looks ok

PR #490, opened 2022-12-13
Bom ut 4

lewisrenfrew:
- not much extra functionality but shows the proposed changes now at least


richardiphillips (approved):
I don't really understand this but looks clever

lewisrenfrew:
> I don't really understand this but looks clever

Clever... Or massively overcomplicated? I'll go over all the bits at the next show & tell and we can decide.

PR #501, opened 2022-12-23
prohibit ordering non live parts

lewisrenfrew:
- is this sensible? Should this only apply to stock controlled parts? @andrewfraser73


richardiphillips (approved):
yeah i think it should be for all parts

PR #504, opened 2023-01-06
bom details

andrewfraser73:
show a popup of bom change details on change requests could also form the basis of a crf print.

would have liked to a sort of nested datagrid view but datagrids within datagrids too ugly.


richardiphillips (approved):
looks believable

PR #507, opened 2023-01-09
Adv func search bom

lewisrenfrew:
- allows you to search for a part on the bom via the tree ui just like the old form
- works like ctrl + f, so flicks through matching results one by one if there is more than one
- I also made the page scroll to the current search result, although its  hacky/not very react-y and there might be a better way


RossStewart (approved):
Looks good to me!

PR #510, opened 2023-01-11
Adv func copy bom

lewisrenfrew:
- pre-support wip commit, but most of the bits are in place


andrewfraser73 (approved):
in general this will work but there might be a scenario with an approved change request where this falls over

PR #514, opened 2023-01-17
More bom ut

lewisrenfrew:
- various logic fixes, hopefully. will show and tell all this tomorrow if we have time
- delete bom advanced function, which is really a 'Delete All From Bom' function as what it actually does is propos a change to remove all details from a given bom


richardiphillips (approved):
looks ok to me but i'm no expert

PR #515, opened 2023-01-17
some more stuff about pcas boards

richardiphillips:


src/Facade/Services/CircuitBoardFacadeService.cs
```diff
@@ -5,31 +5,100 @@
...
         public IResult<CircuitBoardResource> UpdateBoardComponents(
-            int id,
-            CircuitBoardResource updateResource,
+            string id,
+            CircuitBoardComponentsUpdateResource updateResource,
             IEnumerable<string> privileges = null)
         {
-            throw new NotImplementedException();
+            var pcasChange = this.pcasChangeRepository.FindBy(
+                a => a.BoardCode == id && a.RevisionCode == updateResource.ChangeRequestRevisionCode
```
lewisrenfrew, line 56: Does this pcas_change need to be in any specific state? For example is it possible this FindBy could fish out a change that's already LIVE or CANCEL'd to track this update against?

lewisrenfrew (approved):

src/Facade/Services/CircuitBoardFacadeService.cs
richardiphillips: good point

PR #517, opened 2023-01-20
Bom ut final bits

lewisrenfrew:


src/Domain.LinnApps/Boms/BomChangeService.cs
```diff
@@ -293,12 +296,15 @@ public void CopyBom(string srcPartNumber, string destBomPartNumber, int changedB
...
             }
 
+            // todo - hardcoded PROPOS? should this use the change in question's state?
             this.bomPack.CopyBom(srcPartNumber, destBom.BomId, change.ChangeId, "PROPOS", "O");
         }
 
         public void DeleteAllFromBom(string bomName, int crfNumber, int changedBy)
         {
             var bom = this.bomRepository.FindBy(x => x.BomName == bomName);
+            // todo - below is wrong
```
lewisrenfrew, line 306: just realised I haven't done these todos - will do before I merge

src/Domain.LinnApps/Boms/BomChangeService.cs
```diff
@@ -269,37 +272,41 @@ public BomTreeNode CreateBomChanges(BomTreeNode tree, int changeRequestNumber, i
...
-                                 DateEntered = DateTime.Today,
-                                 EnteredById = changedBy,
-                                 ChangeState = "PROPOS",
-                                 PartNumber = destBom.BomName,
-                                 Comments = "BOM_UT",
-                                 PcasChange = "N"
-                             };
-                this.bomChangeRepository.Add(change);
+                var detail = this.bomDetailRepository.FindById(child.DetailId);
+                if (!detail.DeleteChangeId.HasValue && new[] { "PROPOS", "ACCEPT", "LIVE" }
```
andrewfraser73, line 286: what if the changes were added by this change request will this function just delete them or will it mark them as deleted by the same change request i.e. same add and delete change id?

andrewfraser73 (approved):
think it's fine but if you need me to try and break it at some point let me know
think there are some nuances around deleting lines you've proposed or editing approved lines but this seems like a step forward

src/Domain.LinnApps/Boms/BomChangeService.cs
lewisrenfrew: yeah good point - as it stands probably the latter which probably isn't right. It should just make them disappear really shouldn't it?

PR #519, opened 2023-01-20
Wildsearch

andrewfraser73:
Ability to search change requests using wildcards either * or % work
The following work
- TESTASS3
- TESTASS*
- *3
- *LEWIS*

Would have liked a LINQ like that worked with EF but Regex not supported and hence clunky if to cope with these conditions and can't cope with something like PCAS */L1* but that's an edge case

also added a filter with ability to show recent live changes but will need Jacki to see if these values are useful or not

### DIFF
--- a/src/Facade/Services/ChangeRequestFacadeService.cs
+++ b/src/Facade/Services/ChangeRequestFacadeService.cs
@@ -4,7 +4,6 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
-    using System.Text.RegularExpressions;
 
     using Linn.Common.Domain.Exceptions;
     using Linn.Common.Facade;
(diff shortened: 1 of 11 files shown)


src/Facade/Services/ChangeRequestFacadeService.cs
```diff
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
+    using System.Text.RegularExpressions;
```
richardiphillips, line 7: not being used if you've abandoned regex?

src/Facade/Services/ChangeRequestFacadeService.cs
```diff
@@ -195,6 +196,59 @@ public IResult<IEnumerable<ChangeRequestResource>> GetChangeRequestsRelevantToBo
...
 
+        public IResult<IEnumerable<ChangeRequestResource>> SearchChangeRequests(
+            string searchTerm,
+            bool? outstanding, 
+            int? lastMonths,
+            IEnumerable<string> privileges = null)
+        {
+            var changeRequests = new List<ChangeRequest>();
+
+            var fromDate = (lastMonths == null)
```
richardiphillips, line 207: could move into relevant part of IF statement. Also the hard coded date will date quickly. probably better to be relative to today

src/Facade/Services/ChangeRequestFacadeService.cs
```diff
@@ -195,6 +196,59 @@ public IResult<IEnumerable<ChangeRequestResource>> GetChangeRequestsRelevantToBo
...
+                               ? new DateTime(2020, 1, 1)
+                               : DateTime.Now.AddMonths(-1 * (int)lastMonths);
+            var inclLive = ((outstanding == false) ? "LIVE" : "JUSTOUTSTANDING");
```
richardiphillips, line 210: feels like too many brackets

src/Facade/Services/ChangeRequestFacadeService.cs
```diff
@@ -195,6 +196,59 @@ public IResult<IEnumerable<ChangeRequestResource>> GetChangeRequestsRelevantToBo
...
+
+            var newPartNumber = searchTerm.Trim().ToUpper().Replace('%', '*');
+            var partSearch = newPartNumber.Split('*');
+
+            // this big if is just because Linq/EF/Oracle doesn't do a LIKE
+            // supports IC*, *3, *LEWIS*, PCAS*L1R1 but not multiple * e.g. PCAS */L1*
+            if (string.IsNullOrEmpty(newPartNumber))
+            {
+                var a = this.repository.FilterBy(
```
richardiphillips, line 219: is this just testing code that can be removed?

richardiphillips (approved):

src/Facade/Services/ChangeRequestFacadeService.cs
```diff
@@ -253,5 +307,7 @@ protected override void DeleteOrObsoleteResource(ChangeRequest entity, IEnumerab
         {
             throw new NotImplementedException();
         }
+
+
```
lewisrenfrew, line 311: blankity blank

PR #522, opened 2023-01-24
catch mimekit address error

richardiphillips:


src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -636,9 +636,16 @@ public ProcessResult EmailMultiplePurchaseOrders(IList<int> orderNumbers, int us
...
+                    try
+                    {
+                        this.SendOrderPdfEmail(html, supplierContactEmail, copyToSelf, userNumber, order, null);
+                        text += $"Order {orderNumber} emailed successfully to {supplierContactEmail}\n";
+                        success++;
+                    }
+                    catch (MimeKit.ParseException exception)
+                    {
+                        text += $"Order {orderNumber} to {supplierContactEmail} failed. {exception.Message}. \n";
+                    }
```
lewisrenfrew, line 648: you could also just add a generic Exception catch after this more specific one in case any unforseen exceptions raise their heads?

lewisrenfrew (approved):

lewisrenfrew (approved):
nice. could also log the process result's text to syslog so we can see it too?

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
richardiphillips: good idea. i was swithering whether to catch more than that

richardiphillips:
> 

literally just did that too

PR #524, opened 2023-01-24
fill part and supplier id search boxes when supplied in query string

richardiphillips:


lewisrenfrew:
seems fine, although id be tempted just to set these as initial values in the useState calls for the relevant state fields if that works just as well

lewisrenfrew (approved):

PR #526, opened 2023-01-26
Assembly

andrewfraser73:
Ability to create and view board changes
there is a link through to the circuit board app that goes to the URL but doesn't quite work yet


src/Service.Host/client/src/components/ChangeRequests/ChangeTypes/BoardChange.js
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return [];
+    };
+
+    return (
+        <>
+            {creating ? (
+                <>
+                    <Grid item xs={4}>
+                        <Typeahead
```
lewisrenfrew, line 82: Could try the <Search /> component instead of the Typeahead - think we should slowly move towards deprecating Typeahead if we can since search is a little easier to use

lewisrenfrew (approved):
🎉🎉
🎂🎂

src/Service.Host/client/src/components/ChangeRequests/ChangeTypes/BoardChange.js
```diff
@@ -0,0 +1,171 @@
...
+                            value={item?.revisionCode}
+                            label="RevisionCode"
+                            propertyName="revisionCode"
+                            disabled
+                        />
+                    </Grid>
+                    <Grid item xs={12}>
+                        <Button
+                            onClick={() => {
+                                history.push(`/purchasing/boms/board/${item?.boardCode}`);
```
richardiphillips, line 139: boards rather than board should fix it I think

PR #533, opened 2023-02-01
Even more bom ut

lewisrenfrew:
- mainly just a big refactor of the BomChangeService (which github has handily hidden since the diff is so big - make sure you expand that one out if you are reviewing to see what I actually changed)
 - the general idea was to split out the messy spaghetti of the main 'ProcessTreeUpdate' method  into hopefully more understandable smaller methods. One for traversing, one for processing each sub assembly, and then specialised ones for the different kinds of updates we do to bom tree nodes (add, delete, undo delete, replace)
 - I also removed the added complexity of keeping the tree (in memory) up to date and returning that to the client. Now the state of the tree in the db is the single source of truth, and I query the database after any updates commit to guarantee the most-up-to-date tree is returned to the client - trade off is I have to hit the db again obviosuly to rebuild the tree, which makes the PostBomTree request a bit slower.
 - all kinds of bug fixes

### DIFF
--- a/src/Domain.LinnApps/Boms/BomChangeService.cs
+++ b/src/Domain.LinnApps/Boms/BomChangeService.cs
@@ -250,6 +250,7 @@
             }
 
             node.DeleteChangeDocumentNumber = null;
+            node.ChangeState = "LIVE"; // todo - check this is right - add to test if so
             this.bomDetailRepository.FindById(int.Parse(node.Id)).DeleteChangeId = null;
         }
 
@@ -277,7 +278,6 @@
 
             replacedDetail.DeleteChangeId = change.ChangeId;
             replacedDetail.DeleteReplaceSeq = seq + 1;
-            replacedDetail.ChangeState = change.ChangeState;
         }
 
         private Bom GetOrCreateBom(string name)
--- a/src/Domain.LinnApps/Boms/BomChangeService.cs
+++ b/src/Domain.LinnApps/Boms/BomChangeService.cs
@@ -164,32 +164,7 @@
 
         private void AddNode(BomTreeNode node, BomChange change, int? replacementSeq)
         {
-            var part = this.partRepository.FindBy(p => p.PartNumber == node.Name);
-
-            if (part == null)
-            {
-                throw new ItemNotFoundException($"Invalid Part Number: {node.Name} on Assembly: BOM");
-            }
-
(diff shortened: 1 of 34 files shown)


src/Service.Host/client/src/components/BomUtility/BomUtility.js
```diff
@@ -119,14 +119,19 @@ function BomUtility() {
...
             crNumber && contextMenu === null
                 ? {
                       mouseX: e.clientX + 2,
                       mouseY: e.clientY - 6,
                       part: target.innerText,
-                      canReplace: Number(crNumber) !== detail.addChangeDocumentNumber,
+                      canDelete: !detail.deleteReplaceSeq,
+                      canReplace:
+                          Number(crNumber) !== detail.addChangeDocumentNumber &&
+                          !detail.addReplaceSeq &&
```
andrewfraser73, line 132: does the !addReplaceSeq prevent chaining of change requests i.e. replace a component on one change request then change that change to another part number in another one.  not common but sometimes Jacki might do crazy stuff like that (worth asking)

andrewfraser73 (approved):
seems ok one tiny query about chaining of changes

src/Service.Host/client/src/components/BomUtility/BomUtility.js
lewisrenfrew: Yeah - as it stands you can delete or replace a change that is involved in a deletion/replacement on a different change request

src/Domain.LinnApps/Boms/BomChangeService.cs
```diff
@@ -345,13 +103,207 @@ public void DeleteAllFromBom(string bomName, int crfNumber, int changedBy)
...
+        {
+            // todo - is this correct? - if so add a unit test to enshrine it
+            if (node.DeleteChangeDocumentNumber.GetValueOrDefault() != changeRequestNumber)
+            {
+                throw new InvalidBomChangeException(
+                    $"cannot undo delete of {node.Name} - it was deleted on a different change request.");
+            }
+
+            node.DeleteChangeDocumentNumber = null;
+            node.ChangeState = "LIVE"; // todo - check this is right - add to test if so
```
lewisrenfrew, line 253: @andrewfraser73 any thoughts on this todo? My thinking is if you undo a delete node then that node needs to go back to being 'LIVE' - can you think of any situation where that won't be the case?

PR #536, opened 2023-02-06
Undo Changes

andrewfraser73:
Jacki likes the ability to undo a change in certain circumstances by calling bom_pack or pcas_pack to make a bom_change that is the mirror of a bom_change that has been made live.

limited to only folk by with the admin privilege.


lewisrenfrew (approved):
lgtm

PR #543, opened 2023-02-08
Replace

andrewfraser73:
replace v1 
- the ability to create replace type change requests
- the ability to view old/new parts of replace change requests
- a link to a page that currently merely looks up the change request but will have to do the work to allow picking of parts to globally replace


src/Service.Host/client/src/components/ChangeRequests/CreateChangeRequest.js
```diff
@@ -80,6 +81,9 @@ function CreateChangeRequest() {
         if (item.changeType === 'BOARDEDIT') {
             return item.boardCode?.length > 0 && item.revisionCode?.length > 0;
         }
+        if (item.changeType === 'REPLACE') {
+            return item.oldPartNumber?.length > 0 && item.newPartNumber?.length > 0;
```
lewisrenfrew, line 85: is this just to check these strings exist and are not empty? If so can ditch the ?.length > 0 (I think) since empty strings are falsy anyway

lewisrenfrew (approved):

PR #544, opened 2023-02-09
Explicit

andrewfraser73:
this makes change requests with lots of pcas changes and bom changes load faster
as if you include both bom changes and pcas changes it brings back rows of num of bom changes x num of pcas changes which can be big and slow

by explicitly loading one of them i.e. pcas changes you can split what was one query with lots of rows into two smaller queries.

see this page on explicit loading https://www.entityframeworktutorial.net/EntityFramework4.3/explicit-loading-with-dbcontext.aspx


richardiphillips (approved):
interesting

PR #547, opened 2023-02-13
Fix extraneous bom changes bug

lewisrenfrew:
- fixes a bug where, when adding a subassembly to a bom, an extraneous BomChange was added to the change request for that subassembly (correct behaviour is only one bom change for the bom that is actually changing is created)


andrewfraser73 (approved):
looks ok

PR #553, opened 2023-02-16
Bomhistrep1

lewisrenfrew:
- mostly done, need to check if they still need the 'Std Change' column
- still to add unit tests but will wait till I know I'm on the right track before I do


src/Domain.LinnApps/Reports/BomHistoryReportService.cs
```diff
@@ -0,0 +1,104 @@
...
+                             BomName = g.First().BomName,
+                             DateApplied = g.First().DateApplied,
+                             AppliedBy = g.First().AppliedBy,
+                             DocumentType = g.First().DocumentType,
+                             Operation =
+                                 g.First().Operation
+                                 + (g.Count() > 1 ? Environment.NewLine + g.ElementAt(1).Operation : string.Empty),
+                             PartNumber =
+                                 g.First().PartNumber
+                                 + (g.Count() > 1 ? Environment.NewLine + g.ElementAt(1).PartNumber : string.Empty),
```
richardiphillips, line 93: string interpolation a bit nicer?

src/Domain.LinnApps/Reports/BomHistoryReportService.cs
```diff
@@ -0,0 +1,104 @@
...
+                                                        ? Environment.NewLine + g.ElementAt(1).Qty
+                                                        : string.Empty),
+                             GenerateRequirement = g.First().GenerateRequirement
+                                                   + (g.Count() > 1
+                                                          ? Environment.NewLine + g.ElementAt(1).GenerateRequirement
+                                                          : string.Empty)
+                         });
+        }
+    }
+}
```
richardiphillips, line 104: github complaining about the ending

src/Service.Host/client/src/reducers/index.js
```diff
@@ -132,16 +133,17 @@ const rootReducer = history =>
         automaticPurchaseOrder,
         automaticPurchaseOrderSuggestions,
         batchPurchaseOrderDeliveriesUpload,
+        bomCostReport,
```
richardiphillips, line 136: maybe just a conflict thing but looks like it's putting these back into the wrong order

richardiphillips (approved):
looks good as far as I can understand it

PR #558, opened 2023-02-22
Misc fixes and todos

lewisrenfrew:
- gets much closer to the real bom history... I can maybe explain the problem/ attempted solution at the next show & tell if anyone is interested.
-  general idea is to traverse the bom tree and work out which sub assemblies have relevant BOM_CHANGEs to the bom history in the specified timeframe and then inner join said BOM_CHANGEs with the BOM_HISTORY_VIEW to see how the bom changed
- still no unit tests... they are coming..  I'm just not quite sure I got this right yet
- might be some concepts surrounding delete changes that I'm missing?


andrewfraser73 (approved):
looks okay but probably needs a suite of tests to ensure you are catching all edge cases

PR #563, opened 2023-02-28
Bom hist rep tests and probably fixes too

lewisrenfrew:
- i finally settled on a way of making this work using recursion after about 3 days of bad ideas


andrewfraser73 (approved):
looks okay was worried that some of the data logic needed >= rather than > but on reflection as long as every change gets a fresh time then it should be fine (in the past we sometimes gave things only a date).

PR #572, opened 2023-03-06
v1 with the global replace

andrewfraser73:
does the replace all version of pcas replace by calling the same PL/SQL as the form version
added a couple more links

### DIFF
--- a/src/Proxy/PcasPack.cs
+++ b/src/Proxy/PcasPack.cs
@@ -93,5 +93,72 @@
                 connection.Close();
             }
         }
+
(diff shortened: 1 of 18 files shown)


src/Proxy/PcasPack.cs
```diff
@@ -93,5 +93,72 @@ public void UndoPcasChange(int changeId, int undoneBy)
                 connection.Close();
             }
         }
+
+        public void ReplaceAll(string partNumber, int documentNumber, string changeState, int replacedBy, string newPartNumber)
+        {
+            /*
```
lewisrenfrew, line 99: could delete this comment i suppose

lewisrenfrew (approved):

PR #580, opened 2023-03-14
Purchase order email reminders v2

lewisrenfrew:
- sends an email to suppliers 2 days before the ADVISED_DATE of any outstanding delivery commitments they have with us


richardiphillips (approved):
lg

PR #581, opened 2023-03-15
Disable unit price change when Credit or Returns order

RossStewart:
Disable on Create/within form itself.

### DIFF
--- a/src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
+++ b/src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
@@ -238,7 +238,7 @@
                 this.SetDetailFieldsForCreation(detail, newOrderNumber);
 
                 this.PerformDetailCalculations(
-                    detail, detail, order.ExchangeRate.GetValueOrDefault(1), order.SupplierId, order.DocumentType.Name, creating: true);
+                    detail, detail, order.ExchangeRate.GetValueOrDefault(1), order.SupplierId, creating: true);
 
                 this.AddDeliveryToDetail(order, detail);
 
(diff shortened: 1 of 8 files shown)


lewisrenfrew (approved):
Seems reasonable - although I'd be inclined to restrict at the domain service level if this is an important piece of business logic

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -1055,8 +1056,8 @@ private void PerformDetailCalculations(PurchaseOrderDetail current, PurchaseOrde
 
             current.DetailTotalCurrency = Math.Round(netTotal + current.VatTotalCurrency.Value, 2);
 
-            if (creating || (updated.OrderUnitPriceCurrency == current.OrderUnitPriceCurrency
-                && updated.OurUnitPriceCurrency != current.OurUnitPriceCurrency))
+            if ((creating || (updated.OrderUnitPriceCurrency == current.OrderUnitPriceCurrency
+                && updated.OurUnitPriceCurrency != current.OurUnitPriceCurrency)) && (documentType != "CO" && documentType != "RO"))
```
lewisrenfrew, line 1060: This is a really confusing if statement to read, so maybe I'm getting this wrong.. but looks like 

if documentType is CO, or RO then this if statement is skipped, and OrderUnitPriceCurency is not updated, right?

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -1055,8 +1056,8 @@ private void PerformDetailCalculations(PurchaseOrderDetail current, PurchaseOrde
...
             {
                 // if order price hasn't been overridden but our price has changed, use conv factor
                 current.OrderUnitPriceCurrency = updated.OurUnitPriceCurrency * current.OrderConversionFactor;
```
lewisrenfrew, line 1063: but then is it still possible to get into this 'else if' on the next line? If so that could allow current.OrderUnitPriceCurrency to still be updated? Like I said not too sure how this works so just an obseravtion

src/Domain.LinnApps/PurchaseOrders/PurchaseOrderService.cs
```diff
@@ -238,7 +238,7 @@ public PurchaseOrder CreateOrder(PurchaseOrder order, IEnumerable<string> privil
                 this.SetDetailFieldsForCreation(detail, newOrderNumber);
 
                 this.PerformDetailCalculations(
```
lewisrenfrew, line 240: I think I would have just checked if order.DocumentType was a returns or credit here, and not called this PerformDetailCalculations at all in that case.. 

I think thats a lot clearer and also maybe better? since presumably we don't want anything about the order value to be changed for returns/credit orders anway? E.g. we don't want them changing the qty either as that will change the value too?

lewisrenfrew: @richardiphillips is that right? Should we be stopping the values of these orders changing at all (by restricting changes to price, qty etc completely)

Or just stopping the unit price change?

richardiphillips: i think you can change the qty. might only be returning a few broken ones

PR #582, opened 2023-03-15
some jacki feedback changes

andrewfraser73:
fix a few things for this afternoon's meeting
- move phase in button
- allow editing of live changes

### DIFF
--- a/src/Domain.LinnApps/Boms/BomChangeDetail.cs
+++ b/src/Domain.LinnApps/Boms/BomChangeDetail.cs
@@ -38,6 +38,11 @@
             }
         }
 
+        public BomChangeDetail()
+        {
+            // for json deserialisation        
+        }
+
         public int DetailId { get; set; }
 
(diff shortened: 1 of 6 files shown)


lewisrenfrew (approved):

src/Domain.LinnApps/Boms/BomChangeDetail.cs
```diff
@@ -38,6 +38,11 @@ public BomChangeDetail(BomDetail addDetail, BomDetail deleteDetail, IEnumerable<
             }
         }
 
+        public BomChangeDetail()
+        {
+            // for json deserialisation
```
lewisrenfrew, line 43: How come we need this?

andrewfraser73: not sure someone has changed the shape of the changerequestresource and now it 500s as it doesn't have an empty constructor to deserialise it when you do an edit

PR #586, opened 2023-03-17
delete all the things

lewisrenfrew:
- nice big delete-athon of all the dedicated /export endpoints and associated facade service methods since now content negotiation knows to give you csv if you request a text/csv content type
- need to merge this PR first to get these tests to pass


richardiphillips (approved):
don't really have time to read this but I expect it's right if you want to get it live

PR #604, opened 2023-04-04
Replace6

andrewfraser73:
A tweak for Jacki that when she makes a Board change she can the ability to automatically change old -> new part number

was on post-database-commit before so having to do a facade routine that effectively does an add then a replace as the replace logic needs the change request to exist before it can make bom changes and bom details etc.


richardiphillips (approved):
does all the code have to be in the facade service?

PR #607, opened 2023-04-12
Ensure MaxLength on PurchaseOrder

RossStewart:
Went through PO and made sure we had Max Length on all available fields in an effort to stop overflow. All exception handling looks good to me already but open to suggestions to tighten the form up a bit. 

http://rndsupport.linn.co.uk/scp/tickets.php?id=16492


lewisrenfrew:
> Went through PO and made sure we had Max Length on all available fields in an effort to stop overflow. All exception handling looks good to me already but open to suggestions to tighten the form up a bit.
> 
> http://rndsupport.linn.co.uk/scp/tickets.php?id=16492

the problem highlighted by this ticket wasn't to do with database column constraints, it was a result of the react form not handling errors correctly. I think I fixed it a while ago though so you can close the ticket anyway

lewisrenfrew (approved):

PR #609, opened 2023-04-17
DateRequired must be future date

RossStewart:
http://rndsupport.linn.co.uk/scp/tickets.php?id=17207

### DIFF
--- a/tests/Unit/Domain.LinnApps.Tests/PurchaseOrderReqServiceTests/WhenCreatingOrderFromReqAndCarriageValue.cs
+++ b/tests/Unit/Domain.LinnApps.Tests/PurchaseOrderReqServiceTests/WhenCreatingOrderFromReqAndCarriageValue.cs
@@ -53,7 +53,7 @@
                 PhoneNumber = "+44 1234567780",
                 QuoteRef = "blah",
                 Email = "LC@gmail",
-                DateRequired = 1.January(2033),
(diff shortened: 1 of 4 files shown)


tests/Unit/Domain.LinnApps.Tests/PurchaseOrderReqServiceTests/WhenCreatingOrderFromReqAndCarriageValue.cs
```diff
@@ -53,7 +53,7 @@ public void SetUp()
                 PhoneNumber = "+44 1234567780",
                 QuoteRef = "blah",
                 Email = "LC@gmail",
-                DateRequired = 1.January(2023),
```
richardiphillips, line 56: should make these datetime.now + whatever rather than have tests that will fail in 10 years

richardiphillips (approved):

tests/Unit/Domain.LinnApps.Tests/PurchaseOrderReqServiceTests/WhenCreatingOrderFromReqAndCarriageValue.cs
RossStewart: Yeah good point actually. Makes it foolproof.

PR #620, opened 2023-05-18
Replace7

andrewfraser73:
Jacki wanted ability to edit qty on the add portion of change request replace operations which means changing the back and front end to allow differing quantities.


richardiphillips (approved):
don't really understand it

PR #633, opened 2023-07-20
Support odds and ends

lewisrenfrew:
- use trim() stop whitespaces at start of supplier contacts email address busting the po reminder emails 
- default some dropdowns to No on the Boards Ut
- stop create or update of board with no PCB Part Number on the revision, at jackis request
- remove redundant email address from PO pdfs

### DIFF
--- a/src/Service.Host/client/src/components/boards/BoardTab.js
+++ b/src/Service.Host/client/src/components/boards/BoardTab.js
@@ -136,7 +136,7 @@ BoardTab.defaultProps = {
     idBoard: null,
     defaultPcbNumber: null,
     variantOfBoardCode: null,
-    splitBom: 'N',
+    splitBom: 'Y',
     creating: false
 };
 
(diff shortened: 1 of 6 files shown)


src/Service.Host/client/src/components/boards/BoardTab.js
```diff
@@ -136,7 +136,7 @@ BoardTab.defaultProps = {
     idBoard: null,
     defaultPcbNumber: null,
     variantOfBoardCode: null,
-    splitBom: null,
+    splitBom: 'N',
```
richardiphillips, line 139: did Jacki not want this defaulted to Yes?

src/Service.Host/client/src/components/boards/RevisionTab.js
```diff
@@ -95,7 +87,7 @@ function RevisionTab({
                             <Grid item xs={8} />
                             <Grid item xs={3}>
                                 <Dropdown
-                                    value={revision.splitBom}
+                                    value={revision.splitBom ?? 'No'}
```
richardiphillips, line 90: As above

richardiphillips (approved):

src/Service.Host/client/src/components/boards/RevisionTab.js
lewisrenfrew: yes to both - good catch. if only I could read!

PR #655, opened 2023-10-05
Display correct cancel status

RossStewart:


lewisrenfrew (commented):
whats the difference between item.cancelled and order.cancelled?

RossStewart:
> whats the difference between item.cancelled and order.cancelled?

I believe it's to do with the fact that Order has all its fields initialised when it hits the RECEIVE_NEW_SUGGESTED_PURCHASE_ORDER_VALUES stage whilst Item at that stage has not. Order is also used throughout. 

Happy to look at a different work around though

RossStewart:
> whats the difference between item.cancelled and order.cancelled?

When we get to the RECEIVE_NEW_SUGGESTED_PURCHASE_ORDER_VALUES part of creating a new Returns Order, Item is still looking at the status of the previous Purchase Order that we are looking up and therefore displays a massive CANCELLED beside the title which Anne found confusing. 

This way means that each new PO/CO/RO won't have that big cancelled part beside it unless you actually save and cancel that order. Order is referenced throughout too.

(Sorry for deleted comment, got it mixed round for a second there)

lewisrenfrew:
> > whats the difference between item.cancelled and order.cancelled?
> 
> When we get to the RECEIVE_NEW_SUGGESTED_PURCHASE_ORDER_VALUES part of creating a new Returns Order, Item is still looking at the status of the previous Purchase Order that we are looking up and therefore displays a massive CANCELLED beside the title which Anne found confusing.
> 
> This way means that each new CO/RO that refers to an another cancelled order the title won't have that big cancelled part beside it unless you actually save and cancel that new order. Order is referenced throughout too.
> 
> (Sorry for deleted comment, got it mixed round for a second there)

Yeah I see what you mean / understand how this works now....

Although it might make more sense to just completely clear out any stale purchase order item data when those suggested values arrive (or even the first time the 'create' form is rendered? The purchaseOrderActions should expose a clearItem() function that will dispatch a redux action to clear out the old data, if you wanna try it

I only say the above because having redundant previous state hanging around after we need it is a really common way for bugs to slip into these react forms (e.g if we come back to work on this in a few years time we might naively assume the data hanging around in the 'item' state object contains relevant state when creating a returns order, and not just whatever happened to be hanging around from last time the user looked up a purchase order)

Although if you just wanna make this fix live and make Anne happy thats cool too

lewisrenfrew (approved):

RossStewart:
Had a wee think about this and fully agree with you. Figure it's better to just clear the item on the start of Creating so just updated this with a new commit. Just tested it there and it's working as expected.

PR #669, opened 2023-12-14
Po req nom dept lookup

lewisrenfrew:
- this PR was initially just to improve the req ut nominal/dept lookup http://rndsupport.linn.co.uk/scp/tickets.php?id=18175
- but I ended up updating a bunch of stuff too to get purchasing in line with the most recent comp lib / jest testing /react router / mui stuff
- and changed some Typeaheads to Search's while I was at it
- quite a high opportunity for breaking things with this PR, so worth a decently close look from one or more kind person
- I'll do a bunch of manual testing before I merge too


richardiphillips (approved):
lot's of improvements in there.

PR #683, opened 2024-02-05
BaseOurUnitPrice decimal places

RossStewart:
Updated everywhere we are using the BaseOurUnitPrice to get a total to be 5 decimal places instead of 2 - please check these are all necessary.


richardiphillips (changes requested):
It's only base our unit price and base order unit price that need to change. 

The totals are money amounts and should be 2 dec places. 
Should revert all these changes and find out where the base unit prices are calculated

richardiphillips (approved):

PR #685, opened 2024-03-08
Ticket #18620 - Display correct Doc Type Name in copy

RossStewart:


richardiphillips (approved):

lewisrenfrew (commented):
can't you just do @Model.DocumentType.Name? Why do we need this extra logic in the view?

RossStewart:
> can't you just do @Model.DocumentType.Name? Why do we need this extra logic in the view?

@Model.DocumentType.Name just gives PO/CO/RO, just added this to give it the full name.

lewisrenfrew:
> > can't you just do @Model.DocumentType.Name? Why do we need this extra logic in the view?
> 
> @Model.DocumentType.Name just gives PO/CO/RO, just added this to give it the full name.

sorry, I mean Model.DocumentType.Description, is that not the full name?
https://github.com/linn/purchasing/blob/main/src/Domain.LinnApps/DocumentType.cs

PR #689, opened 2024-04-05
changes to show address Id on supplier ut where tab

andrewfraser73:
show address id in where supplier tab addresses

### DIFF
--- a/src/Service.Host/client/src/components/supplierUtility/tabs/WhereTab.js
+++ b/src/Service.Host/client/src/components/supplierUtility/tabs/WhereTab.js
@@ -16,6 +16,7 @@ import countriesActions from '../../../actions/countriesActions';
 function WhereTab({
     orderAddressId,
     orderFullAddress,
+    invoiceAddressId,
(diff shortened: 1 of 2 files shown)


src/Service.Host/client/src/components/supplierUtility/tabs/WhereTab.js
```diff
@@ -16,6 +16,7 @@ import countriesActions from '../../../actions/countriesActions';
 function WhereTab({
     orderAddressId,
     orderFullAddress,
+    invoiceAddressId,
```
lewisrenfrew, line 19: Was this prop already being passed in from the parent component and just unused?

lewisrenfrew (approved):

src/Service.Host/client/src/components/supplierUtility/tabs/WhereTab.js
andrewfraser73: yeah it seemed like it was intended to be there - i needed it for the supplier display

PR #690, opened 2024-04-24
Po ledger entries pdf links

lewisrenfrew:


src/Facade/ResourceBuilders/PurchaseOrderResourceBuilder.cs
```diff
@@ -129,7 +130,13 @@ public PurchaseOrderResource Build(PurchaseOrder entity, IEnumerable<string> cla
                                             InvoiceRef = e.PlInvoiceRef,
                                             BaseVat = e.BaseVatTotal,
                                             InvoiceDate = e.InvoiceDate.ToString("dd/MM/yyyy"),
-                                            Tref = e.Pltref
+                                            Tref = e.Pltref,
+                                            Links = e.DocumentUploaded == "Y" 
+                                                        ? new LinkResource[1]
```
richardiphillips, line 135: do these need to be on the order resource? Won't most things that need it be requesting the purchase ledger entries anyway?

richardiphillips (approved):

src/Facade/ResourceBuilders/PurchaseOrderResourceBuilder.cs
lewisrenfrew: the post invoices against orders screen fetches the order to render all of the data including the existing trefs posted against that order to stick in the table. It's in this table that I want to conditionally show the download button for a given tref it has a related file on the server. Do you think we should change that page to request the ledger entries from the finance endpoints as opposed to just using the order data?

richardiphillips: inv post gets the order for prices etc but requests the purchase ledger entries separately 
/ledgers/purchase?searchTerm={orderNumber} ?

lewisrenfrew: ah of course. I just saw the request for the order on that screen and naively assumed it was using that data in the table -  I'll just close this pr

PR #696, opened 2024-06-10
update messaging

lewisrenfrew:
- i thought i had to do this to get the resendinf of the order books working - turns out I didn't but thought I might as well keep the updates anyway


richardiphillips (approved):

lewisrenfrew:
I also noticed purchasing.x/dlx are durable=false. Should we change that?

richardiphillips:
> I also noticed purchasing.x/dlx are durable=false. Should we change that?

yeah if it's easy. hasn't caused any problems up till now

lewisrenfrew:
> > I also noticed purchasing.x/dlx are durable=false. Should we change that?
> 
> yeah if it's easy. hasn't caused any problems up till now

I think its just a case of binning the queues/exchanges and then redeploying with the new config - will try merging this PR with durable=True config tonight after hours

PR #700, opened 2024-06-19
fix for ticket

andrewfraser73:
simple fix for an amend price issue i forgot to merge in a few weeks ago


lewisrenfrew (approved):
why not just onChange={handleFieldChange}
it looks like before it was trying to avoid them putting in a value less than 0.01 - but if the lowest bound is changed to zero would you not be just as well removing the constraint (I guess this will still prevent negative numbers but i dont know why you'd ever put in a negative price)

andrewfraser73:
> why not just onChange={handleFieldChange} it looks like before it was trying to avoid them putting in a value less than 0.01 - but if the lowest bound is changed to zero would you not be just as well removing the constraint (I guess this will still prevent negative numbers but i dont know why you'd ever put in a negative price)

i think the problem was them wanting to put in pries less than 0.01 and getting constantly overridden.  this will prevent the negative price case but not the 0 value one

PR #703, opened 2024-07-09
cc lists n stuff

lewisrenfrew:
- allows a cc list to be set up for  a suppliers main order contact emails
- just commenting out some auto emails too at the moment since they are not in use anyway

### DIFF
--- a/tests/Unit/Domain.LinnApps.Tests/SupplierAutoEmailsMailerTests/WhenSendingMonthlyForecastEmail.cs
+++ b/tests/Unit/Domain.LinnApps.Tests/SupplierAutoEmailsMailerTests/WhenSendingMonthlyForecastEmail.cs
@@ -53,17 +53,17 @@
         [Test]
         public void ShouldSendEmailToSupplier()
         {
-            this.EmailService.Received().SendEmail(
-                this.email,
-                this.supplier.Name,
-                null,
-                null,
-                this.supplier.VendorManager.Employee.PhoneListEntry.EmailAddress,
-                this.supplier.VendorManager.Employee.FullName,
-                $"Monthly Forecast - {timestamp}",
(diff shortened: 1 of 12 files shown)


tests/Unit/Domain.LinnApps.Tests/SupplierAutoEmailsMailerTests/WhenSendingMonthlyForecastEmail.cs
```diff
@@ -53,17 +53,17 @@ public void SetUp()
...
-                this.supplier.Name,
-                null,
-                null,
-                this.supplier.VendorManager.Employee.PhoneListEntry.EmailAddress,
-                this.supplier.VendorManager.Employee.FullName,
-                $"Monthly Forecast - {timestamp}",
-                "Please find Monthly order forecast attached",
-                Arg.Is<IEnumerable<Attachment>>(a => a.First().FileName 
-                                                     == $"{this.supplier.SupplierId}_monthly_forecast_{this.timestamp}.csv"));
+            // this.EmailService.Received().SendEmail(
```
andrewfraser73, line 56: is this just commenting out test or does it still do something

andrewfraser73 (approved):
approved but not sure if this PR is commenting out rather than removing some tests

tests/Unit/Domain.LinnApps.Tests/SupplierAutoEmailsMailerTests/WhenSendingMonthlyForecastEmail.cs
lewisrenfrew: it stops some emails being sent, which currently only get sent to a test addres (me) anyway.

PR #707, opened 2024-08-13
Pl restrict nomacc change when bookins

lewisrenfrew:
Tried to clean up a little bit of mess in the PurchaseOrderFacadeService while I was at at - Left a couple of questions against the code changes if anyone has any ideas


src/Facade/Services/PurchaseOrderFacadeService.cs
```diff
@@ -344,50 +327,33 @@ public string GetOrderAsHtml(int orderNumber)
             return this.domainService.GetPurchaseOrderAsHtml(orderNumber);
         }
 
-        public new IResult<PurchaseOrderResource> Add(
```
lewisrenfrew, line 347: Does anyone know why an this 'new' public Add() is hiding the base .Add() here?

src/Facade/Services/PurchaseOrderFacadeService.cs
```diff
@@ -344,50 +327,33 @@ public string GetOrderAsHtml(int orderNumber)
...
-                (PurchaseOrderResource)this.resourceBuilder.Build(order, privileges.ToList()));
-        }
+            this.creditDebitNoteService.CreateDebitOrCreditNoteFromPurchaseOrder(order);
+            this.transactionManager.Commit();
 
-        protected override PurchaseOrder CreateFromResource(
-            PurchaseOrderResource resource,
-            IEnumerable<string> privileges = null)
-        {
-            throw new NotImplementedException();
```
lewisrenfrew, line 384: And similiarly why the usual CreateFromResource as ignored / left unimplemented in favour of a 'new' .Add()?

The only thing I can think of is if whoever wrote this code wanted to prevent a transactionManager.Commit() happening at the end of the request...

richardiphillips: yeah no idea. a lot of this code is a bit weird. good that someone is starting to take a look at it.

richardiphillips: Ross made a good point that there was mini-order complication with this stuff

lewisrenfrew: @RossStewart anything I need to be wary of before I merge this?

PR #708, opened 2024-09-17
Tighten PO errors

RossStewart:


src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
```diff
@@ -152,6 +153,10 @@ function PurchaseOrderUtility({ creating }) {
         itemSelectorHelpers.getItemLoading(state[purchaseOrderDeliveries.item])
     );
 
+    const requestErrors = useSelector(reduxState =>
+        getRequestErrors(reduxState)?.filter(error => error.type !== 'FETCH_ERROR')
+    );
+
```
lewisrenfrew, line 159: out of interest, what so-called 'requestError' does this catch? i.e. which ones aren't error.type == 'FETCH_ERROR'

src/Service.Host/client/src/components/PurchaseOrders/PurchaseOrderUtility.js
```diff
@@ -557,7 +568,7 @@ function PurchaseOrderUtility({ creating }) {
                                     <Grid item xs={12}>
                                         <ErrorCard
                                             errorMessage={
-                                                itemError?.details ?? itemError.statusText
+                                                itemError?.details?.error ?? itemError.statusText
```
lewisrenfrew, line 571: What was the actual problem here? I thought @richardiphillips  hinted at standup that it was just a 500 being returned from an unhandled exception, which we should be catching and handling? + will this change not break error message rendering for every other error that currently renders ok?

lewisrenfrew: is it a 500 just because its an unhandled exception and thats the default behaviour? can't we just catch the exception properly and return a nice result?

lewisrenfrew: So the purchase order form has never been able to display errors correctly?

lewisrenfrew (changes requested):
400 -  BadRequestResults are currently displayed fine by the client (see my screenshot), so  you should leave the client code as it is. Instead, investigate why this particular exception is not being caught and handled correctly and fix it properly. 

The fact that the request is resulting in a 500 implies that the exception is uncaught, and so the application's default uncaught exception handling kicks in, i.e a 500 - internal server error - is returned, which we don't want to happen here.

richardiphillips:
> 400 - BadRequestResults are currently displayed fine by the client (see my screenshot), so you should leave the client code as it is. Instead, investigate why this particular exception is not being caught and handled correctly and fix it properly.
> 
> The fact that the request is resulting in a 500 implies that the exception is uncaught, and so the application's default uncaught exception handling kicks in, i.e a 500 - internal server error - is returned, which we don't want to happen here.

I concur. The problem I saw was not in the client which seems to work fine. It was throwing a 500 unnecessarily. It was something so obvious that I thought I couldn't possibly forget it before I was next on support but unfortunately I instantly forgot what it was.

lewisrenfrew (approved):

PR #709, opened 2024-09-19
add PartDescription to BoardSummary

RossStewart:
Older JM ticket asking for Part Description to be added the Board Quick View

Minor changes and made a change to the Dev view which I'll post in a comment below with changes highlighted. Won't put it on production until approved.


richardiphillips:
no seems like a reasonable solution

richardiphillips (approved):

PR #710, opened 2024-10-15
Ledger period search

OliverMclachlan:
small pr so that i can do a search in cashbook


lewisrenfrew (changes requested):
Why does this need to be in purchasing? seems like more of a finance concern

PR #732, opened 2025-05-20
add year before last to drill down

richardiphillips:


lewisrenfrew (approved):
is that this ticket?
https://rndsupport.linn.co.uk/scp/tickets.php?id=20446
I claimed it but just close if so

PR #736, opened 2025-06-26
allow switching of our qty / our unit price and some updates

richardiphillips:


andrewfraser73 (approved):
seems ok lots of upgrades mixed in there

PR #738, opened 2025-08-11
fix ticket 20875

andrewfraser73:
there was a 7.7k discrepancy because we allow standards to change price to 0 and then when purchasing changed the price to something new the zero price logic kicked in for a scenario where it shouldn't have

ticket https://pchelpdesk.linn.co.uk/scp/tickets.php?id=20876#reply


lewisrenfrew (commented):
empty test?

lewisrenfrew:
> empty test?

scratch that - github is being very slow for me today and wasnt showing the test had been implemented

PR #741, opened 2025-09-17
Expand change request search

lewisrenfrew:
this pr expands the change request search to allow you to search by
- the change requests board code
- by root product, where change requests that either 
  - feature bom changes with parts with the specified root product
  - are for board codes which correspond to parts with the specified root product

+ some bonus refactoring of an interface that had a bunch of methods that were only ever used internal to its implementation


andrewfraser73 (approved):
LGTM - although i remember discussions when we were doing change requests if change_sernos was still being used by Jacki etc.

PR #770, opened 2026-07-21
Schedule forecast pdfs

lewisrenfrew:
- incubating the scheduling code till i am sure it works before I commonise
- just the pdf ones for now to get oracle-win off
- sending to test address and on wednesdays for now so I can see if I receive them all tomorrow (and check they look ok) before turning on for real for Monday mornings


richardiphillips (approved):
should this moved to purchasing2?

